PR #7, opened 2019-06-08
Make drawing errors shareable across threads

rkarp:
Without this, collecting errors from other threads (for example when using rayon) will cause compile errors.


38:
Thanks!

PR #21, opened 2019-08-29
Fix font-family for SVG text

sloshwoven:
This fixes a typo in the "font-family" attribute of SVG text elements.


38:
Thanks for the PR!

PR #23, opened 2019-09-05
Add support for i128 and u128 primitives

Geemili:
Not sure if there are any drawbacks to implementing this.


38:
Thanks for the PR! 

As long as i128 and u128 are stabilized, I can't see any problem with it.

PR #27, opened 2019-09-10
Adding command to open the generated file in chrome window

hajsf:
As we can open the generated drawing using google chrome from command line as below
```bash
C:\Users\hasan.DESKTOP-HU2FQ29\PycharmProjects\rust>start chrome --new-window --app="file:///%cd%/0.png" --incognito --disable-extensions --window-size="640, 640"
```

Where:
- `start chrome` opens google chrome in Windows 10, in Mac it can be replaced by `/Applications/Google\ Chrome.app/Contents/MacOS/Google\` and in Linix it can be replaced by: `google-chrome`

- `--new-window` to open new chrome instant
- `--app` to remove the toolbar
- `%cd%` to get the current path in Windows 10, in Mac and Linux, I think it is `pwd`
- `--incognito` to open in incognito mode
- `--disable-extensions` to disable all extensions
- `--window-size="640, 640"` windows size to match drawing size

I run the above, and got the below:

image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 # Changelog
 
+## Plotters development (?)
+
 ## Plotters 0.2.5 (2019-09-07)
 
 ### Bug Fix
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Plotters development (?)
 
+### Improvement 
+
+- Allowing axis be placed on top or right by setting `right_y_label_area` and `top_x_label_area`
+
 ## Plotters 0.2.5 (2019-09-07)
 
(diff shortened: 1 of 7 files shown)


38:
Hi there, thanks for the good idea. 
But this seems this is not a PR. Would you mind I put your idea into issues. We can discuss there. Thanks :-)

hajsf:
@38 it is ok, I thought to put our first as issues but changed my mind, should the PR be lines of code?

38:
@hajsf No worries. PR should be a request for merge. But your case doesn't actually request to merge anything. Plus, the change set looks a little bit confusing, because it's actually the code I am currently working on. 

All the feature request, bugs, ideas , suggestions should be an issue. :)

PR #32, opened 2019-09-14
Allow for custom svg

MSleepyPanda:
This allows adding custom svg to a document. Opening mainly for discussion.

The svg create is currently reexported, since this now exposes svg types directly from the api and we avoid version hell, be allowing users to use our version. This change should be done with care, since version bumps of svg with breaking changes are also breaking api changes.

### DIFF
--- a/src/drawing/backend_impl/mod.rs
+++ b/src/drawing/backend_impl/mod.rs
@@ -2,6 +2,8 @@
 mod svg;
 #[cfg(feature = "svg")]
 pub use self::svg::SVGBackend;
+#[cfg(feature = "svg")]
+pub use ::svg as svg_types;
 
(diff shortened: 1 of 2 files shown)


38:
Hi @MSleepyPanda , thanks for the PR. Yes, you are right if we need to expose the SVG object to the application, we need to re-export the svg crate as well. 

I like this part. But directly exposing the internal update function isn't consistent with other APIs. 

What is normally done is allowing application pass its an SVG object. like:

```rust
let bg = SVGBackend::with_document(&mut doc); 
// And you should be able to modify the SVG object later
```

I prefer that just because this API consistent with others.  Just like `SVGBackend::with_buffer`, etc...

38:
Finally I have a chance to revisit this issue. And I thought about it a little bit, I think your approach is good. 

Thanks again for the PR.

PR #43, opened 2019-10-04
Change FontLoadError from Rc to Arc for Send

dalance:
I tried to convert from `DrawingAreaErrorKind<DB::ErrorType>` to `failure::Error`, and failed.

```
error[E0277]: `std::rc::Rc<rusttype::Error>` cannot be shared between threads safely
  --> src/plotter.rs:77:54
   |
77 |             .build_ranged(x_min..x_max, y_min..y_max)?;
   |                                                      ^ `std::rc::Rc<rusttype::Error>` cannot be shared between threads safely
   |
   = help: within `plotters::drawing::area::DrawingAreaErrorKind<<T as plotters::drawing::backend::DrawingBackend>::ErrorType>`, the trait `std::marker::Sync` is not implemented for `std::rc::Rc<rusttype::Error>`
   = note: required because it appears within the type `plotters::style::font::ttf::FontError`
   = note: required because it appears within the type `plotters::drawing::backend::DrawingErrorKind<<T as plotters::drawing::backend::DrawingBackend>::ErrorType>`
   = note: required because it appears within the type `plotters::drawing::area::DrawingAreaErrorKind<<T as plotters::drawing::backend::DrawingBackend>::ErrorType>`
   = note: required because of the requirements on the impl of `failure::Fail` for `plotters::drawing::area::DrawingAreaErrorKind<<T as plotters::drawing::backend::DrawingBackend>::ErrorType>`
   = note: required because of the requirements on the impl of `std::convert::From<plotters::drawing::area::DrawingAreaErrorKind<<T as plotters::drawing::backend::DrawingBackend>::ErrorType>>` for `failure::error::Error`
   = note: required by `std::convert::From::from`
```

This seems to be caused by `FontLoadError(Rc<Error>)`.
If `Rc` can be changed to `Arc`, the error will be fixed.

### DIFF
--- a/src/style/font/ttf.rs
+++ b/src/style/font/ttf.rs
@@ -2,8 +2,8 @@ use std::collections::HashMap;
 use std::i32;
 use std::marker::PhantomPinned;
 use std::pin::Pin;
-use std::rc::Rc;
 use std::slice::from_raw_parts;
+use std::sync::Arc;
 use std::sync::Mutex;
 
 use rusttype::{point, Error, Font, Scale};
@@ -20,7 +20,7 @@ type FontResult<T> = Result<T, FontError>;
 pub enum FontError {
     LockError,
     NoSuchFont,
-    FontLoadError(Rc<Error>),
+    FontLoadError(Arc<Error>),
 }
 
 impl std::fmt::Display for FontError {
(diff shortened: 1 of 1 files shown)


38:
Cool, thanks for the fix. In addition to that, I also add a bound in `DrawingBackendError`, which requires the `FontError` to be `Sync + Send`. 

Thanks again for the fix!

PR #50, opened 2019-10-10
Inset feature

serzhiio:
```
        let mut price_chart = ChartBuilder::on(&price_area)
            .x_label_area_size(-15)
            .y_label_area_size(40).inset_y_labels()
            .right_y_label_area_size(40).inset_right_y_labels()
            .build_ranged(from..to, min..max).unwrap()
```
I've found negative values sometimes useful so i enabled i32 instead of u32 in label_size methods.
Hope it helps!


src/chart/builder.rs
```diff
@@ -18,7 +18,8 @@ pub enum LabelAreaPosition {
 /// With the hlep of this object, we can convert a basic drawing area into a chart context, which
 /// allows the high-level chartting API beening used on the drawing area.
 pub struct ChartBuilder<'a, 'b, DB: DrawingBackend> {
-    label_area_size: [u32; 4], // [upper, lower, left, right]
+    label_area_size: [i32; 4], // [upper, lower, left, right]
+    label_area_inset: [bool; 4],
```
38: Just want to discuss with you. Since what I am planning to do is allow developer give a negative label area size, and it will automatically inverts the direction of the labels. Is there any particular reason for explicitly setting this ?

src/chart/builder.rs
```diff
@@ -180,7 +201,20 @@ impl<'a, 'b, DB: DrawingBackend> ChartBuilder<'a, 'b, DB> {
             }
         }
 
-        std::mem::swap(&mut drawing_area, splitted[4].as_mut().unwrap());
+        for (id, (_, size)) in self.label_area_inset.iter().zip(self.label_area_size.iter()).enumerate().filter(|(_,(inset, size))| **inset && **size != 0) {
+            let area = splitted[4].as_ref().unwrap();
+            let (w, h) = area.dim_in_pixel();
+            let mut new_area = match id {
```
38: What I am planning to do is uses the `drawing::area::shrink` function if the sizes are negative to construct the label area. 
I am not sure if we need to define this function, but it seems the `shrink` function also works for this purpose.

We can first clone the parent drawing area and then shrink it.

src/chart/context.rs
```diff
@@ -268,18 +268,25 @@ impl<'a, DB: DrawingBackend, X: Ranged, Y: Ranged> ChartContext<'a, DB, RangedCo
...
+
             if orientation.0 == 0 {
                 x0 = axis_range.start;
                 x1 = axis_range.end;
             } else {
                 y0 = axis_range.start;
                 y1 = axis_range.end;
             }
+
+            if area.is_inset() {
```
38: Just little bit worry about the logic coupling. Since the drawing area is originally designed not to include too many details about chart. 
And later we may have different types of weird coordinate system, and so I am prefer this flag lives in the 2D chart context rather than the generic drawing area.

src/chart/context.rs
```diff
@@ -335,17 +342,25 @@ impl<'a, DB: DrawingBackend, X: Ranged, Y: Ranged> ChartContext<'a, DB, RangedCo
                 };
 
                 area.draw_text(&t, label_style, (text_x, text_y))?;
-
+                // TODO: need some adjustments here
```
38: Actually I am thinking about do the both thing:
1. provide an API for tick_size adjustment 
2. supports negative tick_size
And I am not sure if we can unify the implementation for the inward one.

src/drawing/area.rs
```diff
@@ -106,6 +106,7 @@ pub struct DrawingArea<DB: DrawingBackend, CT: CoordTranslate> {
     backend: Rc<RefCell<DB>>,
     rect: Rect,
     coord: CT,
+    inset: bool,
```
38: Like I said, it's a little bit coupling between the drawing and the chart context here.

src/drawing/area.rs
```diff
@@ -466,9 +476,49 @@ impl<DB: DrawingBackend> DrawingArea<DB, Shift> {
             },
             backend: self.copy_backend_ref(),
             coord: Shift((self.rect.x0, self.rect.y0 + 10 + text_h as i32)),
+            inset: self.inset,
         })
     }
 
+    /// Alter area bounds
+    pub fn alter_diff(self, diff_tl: (i32,i32), diff_rb: (i32,i32)) -> Self {
```
38: Could we just use `shrink` ? It looks a little bit redundant to me , but I may be wrong

src/drawing/backend.rs
```diff
@@ -39,6 +39,8 @@ pub trait BackendStyle {
     fn stroke_width(&self) -> u32 {
         1
     }
+
+    fn title(&self) -> &str { "" }
```
38: And why there's a title here?

38: It seems the title is not actually used by any of the backend.

src/chart/builder.rs
38: Alternatively, we can still use the split, but we can actually discard the one in the middle and use `shrink` to create the plotting area.

src/chart/builder.rs
serzhiio: That was first option i wanted to implement, but then i realized that negative width is usefult itself. It gives more control over drawing areas. It allows to overlay one drawing area onto one with negative size.

src/chart/builder.rs
serzhiio: > 
> 
> What I am planning to do is uses the `drawing::area::shrink` function if the sizes are negative to construct the label area.
> I am not sure if we need to define this function, but it seems the `shrink` function also works for this purpose.
> 
> We can first clone the parent drawing area and then shrink it.
Feel free to copy, change any part. Its important for me that it do the trick.
> 
> 
> Alternatively, we can still use the split, but we can actually discard the one in the middle and use `shrink` to create the plotting area.
As i mentoined earlier negative size is useful itself without labels being inset.

src/chart/context.rs
serzhiio: Since my design needs to be implemented at builder stage it seemed logical to set inset field in `Area `structure. It looks good for me. Pretty generic flag :)

src/chart/context.rs
serzhiio: This part of code i was most confused here i need just mirror all Path Lines coordinates.
Make tw/th = 0 and 0 = tw/th. I faced with some problems when realized there is truncate function. Its good if you know how to make it better.

src/drawing/area.rs
serzhiio: Looks good for me, but you are master. Do as you wish.

src/drawing/area.rs
serzhiio: Sure, show me how :)

src/drawing/backend.rs
serzhiio: Ahhh... i have heavy customized Webgl backend, and when dealing with all Shders, Vertexes and Fragments you need to know what kind of lines you are drawing, sometimes you need to pull through additional color to make some other stuff more smooth. So just treat it like helpful stuff. I'm using it heavily.

38:
Hmm, thanks for explain this. If you don't mind I think I am going to tweak a little bit before I merge it. 

Also, if it's possible, I am very interested in your WebGL backend and it may be really helpful to me. 
I am just asking, it's totally fine if you don't think it's able to open source.

serzhiio:
> 
> 
> Hmm, thanks for explain this. If you don't mind I think I am going to tweak a little bit before I merge it.

Its totally up to you. Make it better please :)
> Also, if it's possible, I am very interested in your WebGL backend and it may be really helpful to me.
> I am just asking, it's totally fine if you don't think it's able to open source.

My WebGl2 mod grew up to whole standalone library. I have no plans making it public so far, but i will consider it in future. Nevertheless if you have any questions abiut just let me know.

PR #60, opened 2019-10-18
Prepare for log base being a parameter

ahcm:
Hi,

what do you think about making the log base a parameter?

Cheers
Andy


38:
Yes, this is definitely useful :)  

Seems you haven't not exposed the base param. So I guess this is one part of the work right? Do you planning keep working on this?

Thanks for the PR
Hao

ahcm:
Sorry, I overlooked your reply. Yes, if you are OK, with this I would continue two work on it.

ahcm:
It might be, that you know better about exposing the param, since it tis your API.

38:
As log base is already added to master, closing this PR. Thanks!

PR #63, opened 2019-10-23
Add support for typestyles like "oblique" or "monospace"

TatriX:
Currently default hardcoded typeface is "Arial", but this font is not
available by default everywhere. This commit allows using typestyles
like "monospace" instead of concrete typefamily, thus using system
defaults.

### DIFF
--- a/src/chart/builder.rs
+++ b/src/chart/builder.rs
@@ -320,14 +320,17 @@ mod test {
         let drawing_area = create_mocked_drawing_area(200, 200, |_| {});
         let mut chart = ChartBuilder::on(&drawing_area);
 
-        chart.caption("This is a test case", ("oblique", 10));
+        chart.caption("This is a test case", (FontFamily::Sans, 10));
 
         assert_eq!(chart.title.as_ref().unwrap().0, "This is a test case");
-        assert_eq!(chart.title.as_ref().unwrap().1.font.get_name(), "oblique");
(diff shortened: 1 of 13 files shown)


TatriX:
I've only tested this on Linux.

38:
Hi,
Thanks for the PR. I confirm it works with OSX as well. I really like the idea and think we definitely should have this. 

But I have a few more notes on that. As you may realized that, we actually use the `Font::name` which is the font-family name when we generate SVG images. 

However, the SVG spec doesn't support "oblique", "italic", "bold" in `font-family` field. 

Generally speaking, I think monospace seems a good description of font name, which refers the default monospace. And otherwise, it seems we could have a default font setting. 

For bold, italic and oblique, all those seems to be a variant of the same font. Thus I think it's better to make another API that allows user choose different variants. 

So to conclude my thoughts, I think it's a good idea in general. And could we only support "default" and "monospace". For other options, I think I should create another task to design a better  API for them.

Thanks again!
Hao

38:
I see, that's cool.  

You mean for the SVG, we can always put the actual font family name to the attribute? But for SVG it seems nice to have some generic font family name. So should it is compatible with SVG and CSS, thus our SVG backend can produce valid output.

TatriX:
For some reason I used `oblique` without actually understanding what it means...
I've added a new type `FontFamily` which should make a bit more sense then previous approach.
Tomorrow I'll take a look at the changes I've made again, as it seems I'm too tired to think properly ;)

PR #64, opened 2019-10-23
Refactor font cache

TatriX:
This change greatly simplifies font caching and removes unsafe
code. It shouldn't affect any functionality. Performance should be
comparable, but of course benchmarks should be run to check that.

NOTE: This commit PR is based on #63 and should be rebased after original one will be merged or rejected.


38:
Ah, this is awesome. I don't know rusttype already have solution to share font data.Plus I saw your RwLock commit, I am very happy to have that.  Thanks :)

38:
Rebased  and merged.

Issue #18, opened 2019-07-08
Make font loading more robust


jkaessens:
When building the font name database, plotters is extremely sensitive to fonts that it cannot load. Plotters bails with "Error: BackendError(FontError(FontLoadError(UnrecognizedFormat)))" even if the font that I'm actually requesting is already loaded.

In my current case and on a system I don't control, there's an OTF formatted font installed in a system directory that breaks plotters for me. If I run it at my local system and temporarily remove the font in question, plotters works fine.

I was expecting that plotters would just skip loading "unloadable" fonts in the look-up table, instead it panics with the above error. Skipping would at least allow plotters to work with the fonts that are actually loadable.

38:
Hi, thanks for feedback. That's an good idea and I am happy to add that.

Also could you please provide a stack backtrace to the panic. Since Plotters should returning error code instead of panicking inside the library. So if you can give me the backtrace it will be very helpful.

Thanks again!

jkaessens:
Hi,
now that you put a finger on it, I realize that it's not actually panicking, it's just my lax error handling that made think it was panicking. Sorry for that!

However, the issue is still there. It's `ChartContext::draw()` returning with the error
```rust
Error: BackendError(FontError(FontLoadError(UnrecognizedFormat)))
```
For reference, it's [Nimbus Sans Regular](https://github.com/ArtifexSoftware/urw-base35-fonts/blob/master/fonts/NimbusSans-Regular.otf) that irritates plotters.

jkaessens:
I could narrow it down to the file [src/chart/mesh.rs:157](https://github.com/38/plotters/blob/40e35e1bb0d5ae4e02a4fbb4b8b4211fc90d1e0d/src/chart/mesh.rs#L157) where a default font "Arial" is loaded. If I replace the font face with something that I do have on my system, i.e. "DejaVu Sans", it works as expected. Specifying a different font for the label does not.

38:
Thanks for digging into it. I guess you are right, and it doesn't support OTF font loading.
The spot you are looking at is for drawing the axis label.

I am going to fix this once I have time. (Currently traveling and no time for coding)
What I am planning to do is allowing customize the default font. (Let me know if this won't help)

Plus I am currently planning to look into the entire font manipulate code since I have realized one of the library I used is leaking memory. I am going to make all those change and cleanup the font manipulating part completely once I have time.

Thanks a lot!

38:
[...]

PR #65, opened 2019-10-24
Replace FontDesc name with new FontFamily type

TatriX:
Currently default hardcoded typeface is "Arial", but this font is not
available by default everywhere.

This PR adds new FontFamily enum which can be used as an
alternative to the font name.

Also see #63

### DIFF
--- a/src/style/font/font_desc.rs
+++ b/src/style/font/font_desc.rs
@@ -45,20 +45,63 @@ impl FontTransform {
 #[derive(Clone)]
 pub struct FontDesc<'a> {
     size: f64,
-    name: &'a str,
+    family: FontFamily<'a>,
     data: FontResult<FontDataInternal>,
     transform: FontTransform,
 }
 
+/// Describes font family
+#[derive(Clone, Copy)]
+pub enum FontFamily<'a> {
+    Serif,
+    SansSerif,
+    Monospace,
+    Name(&'a str),
+}
+
+impl<'a> FontFamily<'a> {
+    pub fn as_str(&self) -> &str {
+        match self {
+            FontFamily::Serif => "serif",
+            FontFamily::SansSerif => "sans-serif",
+            FontFamily::Monospace => "monospace",
+            FontFamily::Name(face) => face,
+        }
+    }
+}
+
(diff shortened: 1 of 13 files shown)


src/style/font/font_desc.rs
```diff
@@ -45,20 +45,63 @@ impl FontTransform {
...
-    name: &'a str,
+    family: FontFamily<'a>,
     data: FontResult<FontDataInternal>,
     transform: FontTransform,
 }
 
+/// Describes font family
+#[derive(Clone, Copy)]
+pub enum FontFamily<'a> {
+    Sans,
```
38: Isn't "sans" same as "sans-serif" but "serif" is different ?
And for CSS it has "serif", "sans-serif" and "monospace". So it seems we should have `Serif` other than `Sans` https://developer.mozilla.org/en-US/docs/Web/CSS/font-family#%3Cgeneric-name%3E   

I am not familiar with fonts. But based on what I learned it seems to be this. I may be wrong on this.

38:
Hi @TatriX ,

Thanks for this PR. I have been thinking about this for a long time!  I think it looks really good. But my question is is "sans" and "sans-serif" the refers to the same thing but "serif" is different ? 

If so, the `FontFamily` enum may need some change. 

I am not very confident on this, I am not an expert on fonts. Please let me know if I was wrong.

Thanks again!

TatriX:
I'm using [css font-family](https://developer.mozilla.org/en-US/docs/Web/CSS/font-family) notation.

TatriX:
Sans Serif:

Serif:

> In typography and lettering, a sans-serif, sans serif, gothic, or simply sans letterform is one that does not have extending features called "serifs" at the end of strokes.

https://en.wikipedia.org/wiki/Sans-serif

src/style/font/font_desc.rs
TatriX: Ah, right. I'm constantly misuse them ;)

38:
Thanks for this amazing work. I was planning this for a long time. 
Plus this should resolve issue #18 

really appreciate this work!

PR #66, opened 2019-10-24
Document with_buffer for BitMapBackend and add helper function

darakian:
First off thanks for the great library. I used it in a project last weekend and I found the `with_buffer` part of the bitmap backed very useful. I wanted to help by adding some documentation and a helper function which aided in using it with the crate `minifb`.

### DIFF
--- a/src/drawing/backend_impl/bitmap.rs
+++ b/src/drawing/backend_impl/bitmap.rs
@@ -146,6 +146,16 @@ impl<'a> BitMapBackend<'a> {
     }
 
     /// Create a new bitmap backend which only lives in-memory
+    ///
+    /// When this is used, the bitmap backend will write to a user provided [u8] array (or Vec<u8>).
+    /// Output is rendered as RGB triples. ie. 
+    /// buff[0] a the red channel for pixel one
+    /// buff[1] is the green channel for pixel one
+    /// buff[2] is the blue channel for pixel one
+    /// and so on
+    ///
+    /// - `buf`: The path to the GIF file to create
+    /// - `dimension`: The size of the image in pixels
     pub fn with_buffer(buf: &'a mut [u8], dimension: (u32, u32)) -> Self {
         Self {
             target: Target::Buffer(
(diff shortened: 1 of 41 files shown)


src/drawing/backend_impl/bitmap.rs
```diff
@@ -165,6 +175,27 @@ impl<'a> BitMapBackend<'a> {
         }
     }
 
+    /// Helper function to convert with_buffer output
```
38: It's useful for sure, but I have a little bit worried about adding this to Plotters.  As you may realized Plotters is a graphing library and any pixel format conversion should be out-of-scope topic in my opinion.  

I actually think changing the pixel format should be a functionality of image manipulation crate. So what you may want to do can be done with the following code as well https://docs.rs/image/0.22.3/image/enum.DynamicImage.html. 

Plotters actually uses this crate internally. To make the conversion convenient, I think exposing an API to borrow the internal image object is better than this function.

38:
Hi @darakian,

Thanks for the PR. The documentation part is really good, but I think the helper function may be a little bit out of our scope. Plotters shouldn't responsible for pixel format conversion, please see my comment and I am really happy to discuss!

Thanks!

src/drawing/backend_impl/bitmap.rs
```diff
@@ -165,6 +175,27 @@ impl<'a> BitMapBackend<'a> {
...
+    ///
+    /// This is a simple helper function which converts RGB triples
+    /// to RGB 32 bit values. Useful when interoprtating with crates
+    /// which expect Vec<32> such as minifb
+    /// Example: 
+    /// ```
+    /// let disp_buff: Vec<u32> = raw_bit_buffer.chunks_exact(3).map(|rgb| rgb_from_u8_to_u32(
```
38: Also, this makes `cargo test` fails.

src/drawing/backend_impl/bitmap.rs
darakian: > Plotters is a graphing library and any pixel format conversion should be out-of-scope topic in my opinion.

Fair enough. As for exposing the internal image, I'm not sure I follow. Do you mean that the  BitMapBacked should expose its internal state? Would it not be better to allow for different write formats at that point?
ex.
```
BitMapBackend::with_buffer(buffer, (width, height), <pixels as triples of u8s/pixels as single u32s/whatever else>)
```

src/drawing/backend_impl/bitmap.rs
darakian: Is it? I had tests fail on me locally, but they were
```
test drawing::area::drawing_area_tests::test_split_grid ... ok
test chart::context::test::test_chart_context ... FAILED
test style::font::ttf::test::test_font_cache ... FAILED
test chart::builder::test::test_caption ... ok
test drawing::backend_impl::bitmap::test_draw_line_out_of_range ... ok
thread panicked while panicking. aborting.
```

I've removed the rgb decode function and the comments and those tests still fail.

38: This seems due to doc test failing. Another threading issue with the test case is fixed this morning. That's mine. 😂

darakian: Gotcha. I looked at my test failure and it was trying to run the code in the comment. My face was like 🤔

Issue #67, opened 2019-10-25
[Feature Request] Improve wasm demo


TatriX:
I'd like to improve wasm demo a bit. In particular I'd like to move all inline styles to already existing css file, extract js code to a separate file, use es6+ in js code, replace setTimeout with requestAnimationFrame and thus remove double buffering example.
 
Also I'd like to make a realtime rendered demo.
If you think this sounds good, I can start working on that ;)

38:
Sure that's awesome! I am really appreciate your contribute. Thank you in advance!

PR #72, opened 2019-10-28
Refactor wasm-demo

TatriX:
Initial commit for the wasm-demo refactor. See #67 

Example works but still need some polishing and documentation.

Any comments appreciated ;)


examples/wasm-demo/www/package-lock.json
```diff
@@ -0,0 +1,5909 @@
+{
```
38: Not really familiar with Web, but if this is generated, I assume this shouldn't be a part of the repository.

TatriX: It's used for reproducible builds: https://stackoverflow.com/questions/44206782/do-i-commit-the-package-lock-json-file-created-by-npm-5
Though some think it can be omitted.

TatriX:
Ok, I think it's ready. Now demo should look nice both on HighDPI screens and mobile, resizing if necessary.

examples/wasm-demo/www/package-lock.json
38: Gotcha, seems `Cargo.lock` stuff. This seems to be able to remove safely in most of the case.

38:
Thanks for the contribution !
Tried the new version example, it looks awesome and better organized. A few things I come up with:

1. The documentation doesn't mentioned `wasm-pack` part and I think we need to have them
2. The old startup script no longer works, the WASM doesn't loading. We need either remove that or fix that.
3. I am not familiar with web - But I think it make sense if we provide an example that doesn't use web-pack and nodejs, and works only  with static content server. 

As I mentioned I am not good at web development at all.  So if you could provide more details, that would be very helpful to me.

Thanks again!

TatriX:
1. Right, I'll add it to the README
Agree with 2 and 3.
I guess I'll make several directories:
  - npm/webpack based as in rustwasm tutorial
  - solution without any bundlers and probably with rust based web server

TatriX:
Now demo can be run both with and without npm. See https://github.com/38/plotters/pull/72/commits/19c855233bae778fc25b0aec45243df9387b5dfa

PR #76, opened 2019-10-30
Parameterize BitMapBackend with Generic Pixel

### LINKED ISSUE #70: [Feature Request] Different Pixel Format for BitMapBackend?
- What is the feature ?
We should support different pixel format for bitmap backend.

- Why do you need the feature ?

If we want to use Plotters to manipulate a framebuffer, the most commonly used frame buffer pixel format is BGRX. Currently we have to draw the RGB buffer and convert the pixel format. 

- What is the use case ?

See above

- Additional Information

#66 This PR is actually trying to address this. But we definitely should have some more clean and nice way to do this inplace

0b01:
Closes #70

### DIFF
--- a/src/drawing/backend_impl/bitmap.rs
+++ b/src/drawing/backend_impl/bitmap.rs
@@ -111,7 +111,13 @@ impl<'a> Buffer<'a> {
 }
 
 /// The backend that drawing a bitmap
-pub struct BitMapBackend<'a, P: 'static + Pixel<Subpixel=u8>> {
+pub struct BitMapBackend<
+    'a,
+    #[cfg(all(not(target_arch = "wasm32"), feature = "image"))]
+    P: 'static + Pixel<Subpixel=u8>,
+    #[cfg(target_arch = "wasm32")]
+    P: 'static
+> {
     _pix: PhantomData<P>,
     /// The path to the image
     #[allow(dead_code)]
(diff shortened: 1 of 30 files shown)


src/drawing/backend_impl/bitmap.rs
```diff
@@ -4,9 +4,9 @@ use std::marker::PhantomData;
 
 #[cfg(all(not(target_arch = "wasm32"), feature = "image"))]
 mod image_encoding_support {
-    pub(super) use image::{ImageBuffer, ImageError, Rgb};
+    pub(super) use image::{ImageBuffer, ImageError, Pixel};
     pub(super) use std::path::Path;
-    pub(super) type BorrowedImage<'a> = ImageBuffer<Rgb<u8>, &'a mut [u8]>;
+    pub(super) type BorrowedImage<'a, P> = ImageBuffer<P, &'a mut [u8]>;
```
0b01: Note: bounds on generic parameters are not enforced in type aliases.

https://github.com/rust-lang/rust/issues/21903

src/drawing/backend_impl/bitmap.rs
```diff
@@ -151,6 +153,7 @@ impl<'a> BitMapBackend<'a> {
         frame_delay: u32,
     ) -> Result<Self, BitMapBackendError> {
         Ok(Self {
+            _pix: PhantomData,
```
38: And I believe this change doesn't support pixel format other than RGB888, unless we change the fast pixel manipulation code as well. See function `blend_rect_fast`, `fill_rect_fast`

PR #79, opened 2019-11-13
Embed Device Support: Pixel format + Incremental rendering tweaking

### LINKED ISSUE #70: [Feature Request] Different Pixel Format for BitMapBackend?

38:
This PR closes #70 

# Background
This change particularly about making Plotters works on embed devices. 

Currently we have RGB pixel format by default. However, some issue mentioned using Plotters to manipulate framebuffer while the most common pixel format for a framebuffer is BGRX. This requires pixel format conversion which waste most of the CPU time. 

I am currently digging into how Plotters works with some embed devices and using framebuffer for realtime data visualization. In this case, it's critical to support the native pixel format for those devices. 

# After the change

- BGRX pixel format would be natively supported, which greatly improves  framebuffer target. (Both real framebuffer and minifb)
- Adding more pixel format would be easy than before
- Introduced new fast alpha blending algorithm, which gives 5x speed up for rectangle blending.
- Tweak the chart layout and make sure incremental rendering works properly.

# Todo list

- [x] Documenting the newly added stuff
- [x] Adding benchmark for BGRX pixel format
- [x] Verify there's no breaking change

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,10 @@
 # Changelog
 
 ## Plotters latest (?)
+### Added
+
+- BitMapBackend now is able to support different pixel format natively. Check our new minifb demo for details.
+- Incremental Rendering by saving the previous chart context into a state and restore it on a different drawing area.
 
 ### Improved
 
--- a/examples/minifb-demo/.gitignore
+++ b/examples/minifb-demo/.gitignore
@@ -1,2 +1,0 @@
-/target
-**/*.rs.bk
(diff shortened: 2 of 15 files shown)


TatriX:
Could you please add a short comment about what is intended purpose of this PR?

38:
> Could you please add a short comment about what is intended purpose of this PR?

Hi @TatriX , Just adding some information about this PR. Please let me know if you have any thoughts.

Issue #82, opened 2019-11-14
[Feature Request] Slice as a discrete coordinate 


38:
### What is the feature ?
Use of slice as a discrete coordinate. We can do 
```rust
.build_ranged(["foo", "bar", "baz"], 0..100)
```

nuald:
My PR #83 may help with that, however the category I've added are not implicitly created from the array and I guess some additional changes will be required.

PR #83, opened 2019-11-15
Boxplot element and category coordinate have been added.

nuald:
Major changes:
- `Quartiles` data object has been added (without outliers support as they are not used so much, but could consume a lot of memory and could have the performance penalty);
- `Boxplot` element has been added (that draws the quartiles);
- `Category` coordinate has been added (as boxplots are usually used for categorical views).

The corresponding example has been added (please note that it support the external data too as I've used it to verify how the dataset with few million rows is going to be calculated and presented). Unfortunately, changing the main `README.md` with an example image and its link would require a commit into the other repo, so it's out of scope of this PR.

Minor changes:
- `TextAlignment` enum has been added as a hint for the rendering engines (used for SVG so far). SVG is rendered on the client side, so the layout based on fonts provided by fontkit is not always correct as the client side may use different fonts. The proper fix could be changing the signature of `draw_text` to use text style and rectangle, so the renderer would calculate the layout by itself, but it's rather a big change that could affect the performance (actually it may improve performance as certain engines don't require pre-calculation of a text block layout).
- Clippy errors and warnings have been addressed or disabled (mostly to reduce the noise as I've used `clippy` for my changes). While some disabled warnings could be properly addressed, that could have the performance penalty and would require careful approach.
- Testing has been added/improved for all the code changes I've introduced. As it's not quite trivial, I've added a notes in `CONTRIBUTING.md` for the future references.


38:
Thanks for the contribution, this is a pretty concrete one! 
I really like all of the things added and just look at the code roughly it looks pretty good. 

Few more questions:

- For the `TextAlignment`

Correct me if I am wrong.
If the size of the text is correctly estimated, there's no effect on the text positioning right. (So bitmap backend can simple discard it)
My understand is this provides a reliable that make sure even text size isn't correctly estimated, we still can make the our anchor point at the right location?
I really like this one, but if this is the case, we should have a better documentation. 

- For category coordinate

I think it's really good, I am thinking about this for a long time. It seems some of your code in this part: 1. get size of the category, 2. convert index to value 3. convert value to index, is used to describe some discrete coord spec and this is exactly what I am going to do when refactoring the discrete trait. By then, I think this coord can be simplified tremendously. 
So I am a little bit hesitating on this one -  SInce the discrete coordinate is to be refactored anyway, after that do we still need keep this or just like have an API issue #82  suggests.  That may be a breaking change then if we remove the category coordinate. 
So do you mind I keep this in version 0.2 and later when version 0.3 is published I would like to use the new discrete trait to replace this one? Thoughts ?

examples/boxplot.rs
```diff
@@ -0,0 +1,222 @@
...
+
+    let mut chart = ChartBuilder::on(&right)
+        .x_label_area_size(40)
+        .y_label_area_size(40)
+        .caption("Horizontal Boxplot", ("sans-serif", 20).into_font())
+        .build_ranged(-30f32..90f32, 0..3)?;
+
+    chart.configure_mesh().line_style_2(&WHITE).draw()?;
+    let area = chart.plotting_area();
+    area.draw(&Boxplot::new_horizontal(1, &quartiles_a))?;
```
38: Just wondering if we can use `draw_series` here, that won't be a very big difference

examples/boxplot.rs
```diff
@@ -0,0 +1,222 @@
...
+        .configure_mesh()
+        .x_desc("Ping, ms")
+        .y_desc(category.name())
+        .y_labels(category.len())
+        .line_style_2(&WHITE)
+        .draw()?;
+
+    for (label, (values, style, offset)) in &series {
+        chart
+            .draw_series(values.iter().map(|x| {
```
38: Seems this gives us some non-determinism in the result (I mean the ordering of series). 
This may make people confusing when run the example and compare the output directly.  
Maybe we can sort the key first?

src/drawing/backend_impl/svg.rs
```diff
@@ -253,9 +255,16 @@ impl<'a> DrawingBackend for SVGBackend<'a> {
         let x0 = pos.0 + offset.0;
         let y0 = pos.1 + offset.1;
 
+        let max_x = (layout.1).0;
+        let (dx, anchor) = match style.alignment {
+            TextAlignment::Left => (0, "start"),
```
38: I feel this `TextAlignment` is more confusing than `TextAnchor`. How do you think about this ? And document this for the fact not all backends agree with fontkits' text size estimation, thus we need to define a invariant point 
Since text alignment may also means we want to align the text on the page.

src/chart/context.rs
```diff
@@ -507,6 +505,9 @@ impl<'a, DB: DrawingBackend, X: Ranged, Y: Ranged> ChartContext<'a, DB, RangedCo
                 tick_size.abs() * 2
             };
 
+        /* All labels are right-aligned. */
+        let label_style = &label_style.alignment(TextAlignment::Right);
```
38: And I believe `DrawingArea::titled` is also benefited from this. (Use middle alignment)

examples/boxplot.rs
nuald: Sure, I'll change it, but wondering why do you prefer `draw_series` over `draw` here (as we're not drawing series here, but rather just one element)?

examples/boxplot.rs
nuald: Definitely, thank you for noticing that.

src/drawing/backend_impl/svg.rs
nuald: TextAnchor sounds good, I like it (I still would prefer to have the rectangle and text alignment within the rectangle, but it's a completely different story). I'll update it.

src/chart/context.rs
nuald: Sure, I'll fix it.

src/drawing/backend_impl/svg.rs
38: But `TextAnchor` is confusing as well. So if you think it's not confusing at all, just make sure documentation describes what it's about. :)

nuald:
> So I am a little bit hesitating on this one - SInce the discrete coordinate is to be refactored anyway, after that do we still need keep this or just like have an API issue #82 suggests. That may be a breaking change then if we remove the category coordinate.
> So do you mind I keep this in version 0.2 and later when version 0.3 is published I would like to use the new discrete trait to replace this one? Thoughts ?

It's completely up to you, I just needed to somehow present the labels and that's the only way I could do it. I'm sure you can do it much better :) I can even remove it from this PR and just left it in my own branch (as I need it right away, and thankfully Cargo can refer to github branches).

src/drawing/backend_impl/svg.rs
nuald: Okay, what about we leave TextAlignment as-is, but rather change the draw_text API to use rectangle instead of (x, y), so the alignment is going to be used properly? I've verified the code, looks like the current implementations could be improved:
- cairo: it's better to use their `text_extents` method to get the proper "text inked" rectangle;
- svg and canvas: don't need "text inked" rectangle calculations at all;
- other backends use `font.draw` method that calculates the rectangle again (!).

However, it's quite a change, and I'd prefer to submit another PR if you don't mind. Also, if you okay with that, do you want full alignments implemented (including vertical), or just horizontal would be enough for now? Surely, we assume one-line text only, multiline text block would require even more alignment options, and could be out of scope of the project completely.

38: Sure, as you might realized, the text positioning code is some legacy convention which can be dated back to the first 300 lines of code in this crate and it causing problems for sure. 
You are more than welcome to make any improvement on this area.  :)

------

And for multi-line text, there's an element for that, called `MultiLineText`, which performing the text layout when it's drawn to backend.

38:
> It's completely up to you, I just needed to somehow present the labels and that's the only way I could do it. I'm sure you can do it much better :) I can even remove it from this PR and just left it in my own branch (as I need it right away, and thankfully Cargo can refer to github branches).

I think I just decided to keep it and replace it when 0.3 released. 

So that's looks fine to me for now :)

src/style/text.rs
```diff
@@ -3,11 +3,26 @@ use super::font::{FontDesc, FontFamily, FontStyle, FontTransform};
 use super::size::{HasDimension, SizeDesc};
 use super::BLACK;
 
+/// The alignment of the text.
+///
+/// Used to determine the invariant (anchor) points for backends
+/// which render the text on the client side.
```
38: Just feel it's better to have more things documented. Such as why this is useful, when we should use this, etc.

src/coord/category.rs
```diff
@@ -0,0 +1,236 @@
...
+
+/// The category element reference (tick).
+pub struct CategoryElementRef<T: PartialEq> {
+    inner: Rc<Vec<T>>,
+    // i32 type is required for the empty ref (having -1 value)
+    idx: i32,
+}
+
+/// The category elements range.
+pub struct CategoryElementsRange<T: PartialEq>(CategoryElementRef<T>, CategoryElementRef<T>);
```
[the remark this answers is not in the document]
nuald: You're right, the initial implementation didn't use reference counting, so I had to use the separate structure to avoid cloning the original vector. But now it's not an issue, and it can be optimized.

src/data/quartiles.rs
```diff
@@ -0,0 +1,127 @@
...
+    /// - `s`: The array of the original values
+    /// - **returns** The newly created quartiles
+    ///
+    /// ```rust
+    /// use plotters::prelude::*;
+    ///
+    /// let quartiles = Quartiles::new(&[7, 15, 36, 39, 40, 41]);
+    /// assert_eq!(quartiles.median(), 37.5);
+    /// ```
+    pub fn new<T: Into<f64> + Copy + PartialOrd>(s: &[T]) -> Self {
```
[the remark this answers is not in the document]
nuald: I'll take a look, I didn't bother with the algorithm so much as was busy trying to make it all work together. I'd appreciate if you share some hints/ideas regarding that though.

src/style/text.rs
nuald: I won't bother too much as I think that the proper implementation will be ready in few weeks, but I'll add few more details, no worries.

src/data/quartiles.rs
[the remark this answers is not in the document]
38: Yep, I don't worry about this too much as well. But median seems doable in linear time complexity and if we allows some approximation, we can even not copying the data chunk. But this is definitely not a big deal.  I am OK with what it looks like currently. Just have a comment here.

src/drawing/backend_impl/bitmap.rs
```diff
@@ -228,6 +228,7 @@ impl<'a> BitMapBackend<'a> {
             .collect()
     }
 
+    #[allow(clippy::cast_ptr_alignment, clippy::many_single_char_names)]
```
38: Just realized recently there's a huge bitmap backend refactoring, that causing some conflict. As long as this file is only fixing clippy issues, so would you mind remove this file from the PR, thus we have less conflict and I want to run clippy again when I get ready to merge :) - Thanks!

src/coord/category.rs
```diff
@@ -8,30 +8,26 @@ use super::{AsRangedCoord, Ranged};
...
-pub struct CategoryElementRef<T: PartialEq> {
-    inner: Rc<Vec<T>>,
     // i32 type is required for the empty ref (having -1 value)
     idx: i32,
 }
 
-/// The category elements range.
-pub struct CategoryElementsRange<T: PartialEq>(CategoryElementRef<T>, CategoryElementRef<T>);
+/// The category elements range
+pub struct CategoryElementsRange<T: PartialEq>(Category<T>, Category<T>);
```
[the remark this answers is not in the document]
nuald: Thank you, I'd appreciate it!

38:
Just Merged the PR. Thanks for the contribution. Really appreciated for this concrete work. :)

wangjiawen2013:
@nuald 
Hi,
I have tested boxplot, it looks good except a little flaw:
image

PR #87, opened 2019-11-22
Text anchor positions have been added.

nuald:
A follow-up PR regarding text alignment. The main change is a new attribute added to the `TextStyle` - the text anchor position that specifies the position of the anchor point (used in `draw_text`) relative to the text itself. I had to dismiss my original idea to use rectangles as there is a `Text` element that is a `Drawable`, therefore need to draw itself using specified points. As a bonus, the current implementation is backward compatible.

Unfortunately, completely get rid of the text size estimation isn't possible yet, therefore there are the use-cases there the text won't be properly aligned yet:
- title as it uses the text height to adjust the drawing area;
- series labels as they need to calculate the bound rectangle;
- multiline text. The current implementation is not quite compatible with the client side rendering as rather the backend should calculate its layout, not the element itself (for example, SVG would need to utilize `<tspan>` tag for that).

However, it doesn't affect the actual visible results so much (unless series labels border is visible) as all of those elements utilize the margins.

Additional notes:
- Now backends do all text transformations, no fonts themselves (as a side effect, the font `draw` code is much simpler now as no offsets calculations are needed). It affect multiline element rendering if rotated, however, it doesn't properly rotate even in the original code;
- `should_draw` check is gone from axis drawing as we can't reliable determine the text bounds, so we utilize clipping for that now (verified and fixed for all backends). Also I guess that users prefer partially visible text than completely hidden. However, it would be nice to have support for mesh margins, so we could put some extra spacing thus avoiding clipping;
- the rendering results changed for all examples, so I guess they need to be updated;
- SVG for MS IE/Current Edge has issues because they don't support all tags from the spec. However, as Edge is switching to Chromium rendering engine, it won't be an issue in near future (I've verified with beta Edge). If it could be an issue, I can try to find some workarounds;
- fixed `draw_pixel` in `bitmap` backend to normalize alpha as the `animation` example send overflowed float values to it.
- `console` example have a gap now between bottom labels and the axis, but it's not because of the miscalculations, but rather because ticks are not drawn there (the `draw_line` function doesn't draw vertical lines with 1 `pixel` width).

### DIFF
--- a/examples/boxplot.rs
+++ b/examples/boxplot.rs
@@ -80,7 +80,6 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
         .configure_mesh()
         .x_desc("Ping, ms")
         .y_desc(category.name())
-        .y_labels(category.len())
         .line_style_2(&WHITE)
         .draw()?;
 
@@ -94,13 +93,14 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
                     .offset(*offset)
             }))?
             .label(label)
-            .legend(move |(x, y)| Rectangle::new([(x - 5, y - 5), (x + 5, y + 5)], style.filled()));
+            .legend(move |(x, y)| Rectangle::new([(x, y - 3), (x + 12, y + 9)], style.filled()));
     }
     chart
         .configure_series_labels()
         .position(SeriesLabelPosition::UpperRight)
         .background_style(WHITE.filled())
         .border_style(&BLACK.mix(0.5))
+        .legend_area_size(22)
         .draw()?;
 
     let drawing_areas = lower.split_evenly((1, 2));
--- a/examples/boxplot.rs
+++ b/examples/boxplot.rs
@@ -80,6 +80,7 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
         .configure_mesh()
         .x_desc("Ping, ms")
         .y_desc(category.name())
+        .y_labels(category.len())
         .line_style_2(&WHITE)
         .draw()?;
 
--- a/examples/console.rs
+++ b/examples/console.rs
@@ -3,6 +3,7 @@ use plotters::drawing::{
     DrawingBackend,
 };
 use plotters::prelude::*;
+use plotters::style::text_anchor::{HPos, VPos};
 use plotters::style::RGBAColor;
 use std::error::Error;
 
@@ -127,10 +128,22 @@ impl DrawingBackend for TextDrawingBackend {
     fn draw_text(
         &mut self,
         text: &str,
-        _style: &TextStyle,
+        style: &TextStyle,
         pos: (i32, i32),
     ) -> Result<(), DrawingErrorKind<Self::ErrorType>> {
-        let offset = pos.1.max(0) * 100 + pos.0.max(0);
+        let (width, height) = self.estimate_text_size(text, &style.font)?;
+        let (width, height) = (width as i32, height as i32);
+        let dx = match style.pos.h_pos {
+            HPos::Left => 0,
+            HPos::Right => -width,
+            HPos::Center => -width / 2,
+        };
+        let dy = match style.pos.v_pos {
+            VPos::Top => 0,
+            VPos::Center => -height / 2,
+            VPos::Bottom => -height,
+        };
+        let offset = (pos.1 + dy).max(0) * 100 + (pos.0 + dx).max(0);
         for (idx, chr) in (offset..).zip(text.chars()) {
             self.0[idx as usize].update(PixelState::Text(chr));
         }
@@ -148,7 +161,6 @@ where
         .margin(1)
         .caption("Sine and Cosine", ("sans-serif", (10).percent_height()))
         .set_label_area_size(LabelAreaPosition::Left, (5i32).percent_width())
-        .set_label_area_size(LabelAreaPosition::Bottom, (10i32).percent_height())
         .set_label_area_size(LabelAreaPosition::Bottom, (10i32).percent_height())
         .build_ranged(-std::f64::consts::PI..std::f64::consts::PI, -1.2..1.2)?;
 
--- a/src/chart/context.rs
+++ b/src/chart/context.rs
@@ -570,16 +570,24 @@ impl<'a, DB: DrawingBackend, X: Ranged, Y: Ranged> ChartContext<'a, DB, RangedCo
                     let (kx0, ky0, kx1, ky1) = if tick_size > 0 {
                         match orientation {
                             (dx, dy) if dx > 0 && dy == 0 => (0, *p - y0, tick_size, *p - y0),
-                            (dx, dy) if dx < 0 && dy == 0 => (xmax - tick_size, *p - y0, xmax, *p - y0),
+                            (dx, dy) if dx < 0 && dy == 0 => {
+                                (xmax - tick_size, *p - y0, xmax, *p - y0)
+                            }
                             (dx, dy) if dx == 0 && dy > 0 => (*p - x0, 0, *p - x0, tick_size),
-                            (dx, dy) if dx == 0 && dy < 0 => (*p - x0, ymax - tick_size, *p - x0, ymax),
+                            (dx, dy) if dx == 0 && dy < 0 => {
+                                (*p - x0, ymax - tick_size, *p - x0, ymax)
+                            }
                             _ => panic!("Bug: Invalid orientation specification"),
                         }
                     } else {
                         match orientation {
-                            (dx, dy) if dx > 0 && dy == 0 => (xmax, *p - y0, xmax + tick_size, *p - y0),
+                            (dx, dy) if dx > 0 && dy == 0 => {
+                                (xmax, *p - y0, xmax + tick_size, *p - y0)
+                            }
                             (dx, dy) if dx < 0 && dy == 0 => (0, *p - y0, -tick_size, *p - y0),
-                            (dx, dy) if dx == 0 && dy > 0 => (*p - x0, ymax, *p - x0, ymax + tick_size),
+                            (dx, dy) if dx == 0 && dy > 0 => {
+                                (*p - x0, ymax, *p - x0, ymax + tick_size)
+                            }
                             (dx, dy) if dx == 0 && dy < 0 => (*p - x0, 0, *p - x0, -tick_size),
                             _ => panic!("Bug: Invalid orientation specification"),
                         }
--- a/src/chart/context.rs
+++ b/src/chart/context.rs
@@ -14,7 +14,8 @@ use crate::coord::{
 use crate::drawing::backend::{BackendCoord, DrawingBackend};
 use crate::drawing::{DrawingArea, DrawingAreaErrorKind};
 use crate::element::{Drawable, DynElement, IntoDynElement, PathElement, PointCollection};
-use crate::style::{AsRelative, FontTransform, ShapeStyle, SizeDesc, TextAlignment, TextStyle};
+use crate::style::text_anchor::{HPos, Pos, VPos};
+use crate::style::{AsRelative, FontTransform, ShapeStyle, SizeDesc, TextStyle};
 
 /// The annotations (such as the label of the series, the legend element, etc)
 /// When a series is drawn onto a drawing area, an series annotation object
@@ -474,23 +475,6 @@ impl<'a, DB: DrawingBackend, X: Ranged, Y: Ranged> ChartContext<'a, DB, RangedCo
         Ok(axis_range)
     }
 
-    fn estimate_right_aligned_label_offset(
-        &self,
-        label_style: &TextStyle,
-        labels: &[(i32, String)],
-    ) -> i32 {
-        labels
-            .iter()
-            .map(|(_, t)| {
-                self.drawing_area
-                    .estimate_text_size(t, &label_style.font)
-                    .unwrap_or((0, 0))
-                    .0
-            })
-            .max()
-            .unwrap_or(0) as i32
-    }
-
     // TODO: consider make this function less complicated
     #[allow(clippy::too_many_arguments)]
     #[allow(clippy::cognitive_complexity)]
@@ -515,28 +499,11 @@ impl<'a, DB: DrawingBackend, X: Ranged, Y: Ranged> ChartContext<'a, DB, RangedCo
         let (tw, th) = area.dim_in_pixel();
 
         /* This is the minimal distance from the axis to the box of the labels */
-        let label_dist =
-            if (tick_size > 0 && orientation.1 > 0) || (tick_size < 0 && orientation.1 < 0) {
-                0
-            } else {
-                tick_size.abs() * 2
-            };
-
-        /* All labels are right-aligned. */
-        let label_style = &label_style.alignment(TextAlignment::Right);
+        let label_dist = tick_size.abs() * 2;
 
         /* Draw the axis and get the axis range so that we can do further label
          * and tick mark drawing */
         let axis_range = self.draw_axis(area, axis_style, orientation, tick_size < 0)?;
-
-        /* If the label area is on the right hand side, we should enable the right aligned
-         * layout, thus in this case we need to estimate the right most position when all
-         * the labels are right aligned */
-        let right_alignment = if orientation.0 > 0 && orientation.1 == 0 {
-            self.estimate_right_aligned_label_offset(label_style, labels)
-        } else {
-            0
-        };
 
         /* Then we need to draw the tick mark and the label */
         for (p, t) in labels {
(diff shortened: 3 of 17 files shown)


src/drawing/backend_impl/bitmap.rs
```diff
@@ -231,6 +231,11 @@ pub trait PixelFormat: Sized {
                         *buf.get_unchecked_mut(base + idx) = Self::byte_at(r, g, b, 0, idx);
                     }
                 } else {
+                    let alpha = alpha.min(1.0).max(0.0);
```
38: Just wondering isn't this already checked to be less than 1.0 already? There might be a overflow, but it seems what we should do is changing the if condition from `1.0` to `255.0 / 256.0`?

examples/console.rs
```diff
@@ -127,10 +128,22 @@ impl DrawingBackend for TextDrawingBackend {
...
+        let (width, height) = (width as i32, height as i32);
+        let dx = match style.pos.h_pos {
+            HPos::Left => 0,
+            HPos::Right => -width,
+            HPos::Center => -width / 2,
+        };
+        let dy = match style.pos.v_pos {
+            VPos::Top => 0,
+            VPos::Center => -height / 2,
+            VPos::Bottom => -height,
```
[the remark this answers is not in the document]
38: I went through the code, it seems the new code doesn't really guarantee the text isn't in the bouncing box.  Since the decedent part will be always out of box, and this should be avoid.

src/chart/context.rs
```diff
@@ -549,81 +516,70 @@ impl<'a, DB: DrawingBackend, X: Ranged, Y: Ranged> ChartContext<'a, DB, RangedCo
...
-            let should_draw = if orientation.0 == 0 {
-                cx >= 0 && cx + label_offset + w as i32 / 2 <= tw as i32
+            let (text_x, text_y) = if orientation.0 == 0 {
+                (cx + label_offset, cy)
             } else {
-                cy >= 0 && cy + label_offset + h as i32 / 2 <= th as i32
+                (cx, cy + label_offset)
             };
 
-            if should_draw {
```
38: And one reason for this `should_draw`check is we need to avoid the tag is drawn out of it's own area, but if we only relies on the backend, a super long label may overlap with other drawing area on the right side (for example). And I think we should avoid this

src/style/font/ttf.rs
```diff
@@ -159,31 +159,25 @@ impl FontData for FontDataInternal {
...
         let scale = Scale::uniform(size as f32);
         let mut result = Ok(());
         let font = &self.0;
 
-        let base_x = x + trans.offset(layout).0;
-        let base_y = y + trans.offset(layout).1;
-
         for g in font.layout(text, scale, point(0.0, 0.0)) {
             if let Some(rect) = g.pixel_bounding_box() {
-                let x0 = rect.min.x;
```
38: And I believe this is one of the reason why the descender is out of the box now.

examples/boxplot.rs
```diff
@@ -94,13 +93,14 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
                     .offset(*offset)
             }))?
             .label(label)
-            .legend(move |(x, y)| Rectangle::new([(x - 5, y - 5), (x + 5, y + 5)], style.filled()));
+            .legend(move |(x, y)| Rectangle::new([(x, y - 3), (x + 12, y + 9)], style.filled()));
```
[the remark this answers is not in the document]
nuald: Legend is drawn correctly, I've just changed the rectangle location to be aligned more with the text (original change had a misalignment already, but I hoped that it could be fixed with the proper text rendering, but as I mentioned above, the legends rendering is not easy to fix and out of scope of this PR).

src/style/font/ttf.rs
nuald: Yes, it's a regression bug I've introduced, the up-to-dated change has the proper calculations.

src/chart/context.rs
nuald: It could be confusing to users - if super long label is not drawn at all, they may think it's a bug in their code. I think the proper fix would be implement clipping in the drawing area, but I'm not sure how to properly do it with the current architecture (as drawing backend is not supposed to know about the drawing area). I guess it's possible to add a new API call for the backend like `set_clipping_region`, but it's rather high-level change that I won't risk to add without your explicit approval.

nuald:
Few additional notes:
- Is there any specific reason for not using `draw_pixel` for client-side backends (SVG and Canvas)? I've verified, just removing `draw_text` from those backends makes the resulting images are properly rendered with the server-side font. Surely, it has performance penatly and pixels are visible if the image is scaled, but if users require better text positioning, it would be nice to have an option to use the server-side rendering.
- You may notice that cairo has issues with `draw_text` on Windows (`_win32_scaled_font_set_world_transform` error with _helpful_ description "error occurred in the Windows Graphics Device Interface" that comes from WinAPI itself), so with the new change full test suite is going to fail quite often there (as a lot of `draw_text` calls are used). I've dig it a little, but couldn't find an easy fix:
  - it's not because tests are multithreaded (I've tested a global mutex for all font/text related operations);
  - setting fonts explicitly instead of auto-selecting doesn't help either (and it could be an actual bug in the cairo-win32-font as it doesn't directly use the font from the context, but looks like it creates additional ones and have it messed up sometimes);
  - the proper fix could be avoiding _toy_ functions and using proper FreeType fonts loaded by font-kit (in that case cairo-win32-font won't be probably even used), but it requires some additional work which I'm not sure anybody would care about :)

38:
>     * Is there any specific reason for not using `draw_pixel` for client-side backends (SVG and Canvas)?

For this one, I think it's an important feature to use the client-side rendered text. Otherwise, some of the client is very slow and we just lose the reason for using SVG image format at all. But yes, we actually can render a bitmap image and blit it onto the SVG image but I think this is another story. 

> You may notice that cairo has issues with draw_text on Windows 
I am not sure, since I don't really have a Windows machine on my hand. But I am not sure if there is anybody who really care about the test case fails on Windows. It's not bothering me at all, but I would may want to make sure that it's something really related to cairo-rs bug specifically on Window. If this is the case I  think I can make it a low priority task.

38:
image
Another regression I realized is the tick labels on the right side, seems not properly aligned.

src/drawing/backend_impl/bitmap.rs
```diff
@@ -231,7 +231,7 @@ pub trait PixelFormat: Sized {
                         *buf.get_unchecked_mut(base + idx) = Self::byte_at(r, g, b, 0, idx);
                     }
                 } else {
-                    let alpha = alpha.min(1.0).max(0.0);
+                    let alpha = alpha.min(255.0 / 256.0).max(0.0);
```
38: I believe we can juse use if alpha >= 255.0 / 256.0 as the check for if we need to directly override the pixel.

src/chart/context.rs
```diff
@@ -549,57 +516,54 @@ impl<'a, DB: DrawingBackend, X: Ranged, Y: Ranged> ChartContext<'a, DB, RangedCo
...
-                .estimate_text_size(&t, &label_style.font)
-                .unwrap_or((0, 0));
-
-            let (cx, cy) = if tick_size >= 0 {
+            let (cx, cy, h_pos, v_pos) = if tick_size >= 0 {
                 match orientation {
                     // Right
                     (dx, dy) if dx > 0 && dy == 0 => {
-                        (right_alignment + label_dist - w as i32, *p - y0)
+                        (label_dist, *p - y0, HPos::Left, VPos::Center)
```
38: Could we use right alignment here?

src/chart/context.rs
38: Ok, I think your comment make sense. I think I am Ok with this . But I don't think we don't want the backends aware of any high-level details. Since I feel we should keep high-level and low-level API separate. `set_clipping_region` solution might be some solution, but  I think it needs some amount of work to do and I don't think it's worth us to dig into this.

src/drawing/backend_impl/svg.rs
```diff
@@ -247,24 +248,24 @@ impl<'a> DrawingBackend for SVGBackend<'a> {
...
-            TextAlignment::Right => (max_x, "end"),
-            TextAlignment::Center => (max_x / 2, "middle"),
+        let (x0, y0) = pos;
+        let text_anchor = match style.pos.h_pos {
+            HPos::Left => "start",
+            HPos::Right => "end",
+            HPos::Center => "middle",
         };
+        let dominant_baseline = match style.pos.v_pos {
+            VPos::Top => "hanging",
```
38: Just tried with `eog`, which is `librsvg` based SVG rendering, it seems it doesn't support this. So could we think about not supporting the vertical anchor point for now, because I think we should generate SVG that renders properly in most of the clients.

38: Or at least, we don't rely on `dominant_baseline` for a while until this is widely supported.

src/drawing/backend.rs
```diff
@@ -191,15 +192,37 @@ pub trait DrawingBackend: Sized {
...
+        let width = (max_x - min_x) as i32;
+        let height = (max_y - min_y) as i32;
+        let dx = match style.pos.h_pos {
+            HPos::Left => 0,
+            HPos::Right => -width,
+            HPos::Center => -width / 2,
+        };
+        let dy = match style.pos.v_pos {
+            VPos::Top => height,
+            VPos::Center => height / 2,
```
38: I believe you have changed the `Font::draw` to use the baseline as coordinate, this is perfect. But if this is the case, we may just have a very small offset when we use the middle anchor point. I suspect this is related to the small offset in all the Y-axis label
image

src/style/font/font_desc.rs
```diff
@@ -255,7 +256,7 @@ impl<'a> FontDesc<'a> {
         TextStyle {
             font: self.clone(),
             color: color.to_rgba(),
-            alignment: TextAlignment::Left,
+            pos: Pos::default(),
```
38: Not about this one, but I believe after the change, we can get rid of `FontTransform::offset` right? Because we have better model that rotating the text based on the anchor point as the center.

src/drawing/backend_impl/bitmap.rs
nuald: Could you please elaborate? The then-branch above has the check for overriding the pixel (alpha >= 1.0), do I miss something here?

src/chart/context.rs
nuald: The labels on the right side of the chart use the left alignment as before, nothing changed here. The name `right_alignment` was quite confusing as it was used in a meaning "the max width of the labels" (therefore the calculations produced the proper offset for all labels being on the same vertical line with the distance of `label_dist` from the axis).

src/drawing/backend_impl/svg.rs
nuald: Thanks, I'll add the tests and try to find a solution.

src/style/font/font_desc.rs
nuald: Yes, I'll remove it.

nuald:
You're absolutely right that there are some issues with the middle anchor points, but it's not a regression bug, it's something unexpected I couldn't see before until you pointed it to me (but maybe you know the reason). I've modified the tests to draw the original layout_box returned by the `rusttype` font itself (as red circles for min and max `pixel_bounding_box`):

layout_box

As you could see the `x` bounds are correct, but `y` are not. I'm going to research it, and I hope it's not rusttype bug.

nuald:
I've re-verified `rusttype` bbox from the scratch, and while it does have minor differences from the actual pixels it draws, the problem was in my y-values calculations, I've fixed it.

SVG has been fixed too to use `ex`-based lengths with `dy` instead of `dominant_baseline` (and actually, it's improved the accuracy a little). Unfortunately, it affects titles even more now, but I doubt we can do anything with it without changing/adding some API.

src/chart/context.rs
38: I believe there's a change here. Since before the change, we just make sure the right aligned for all the labels. 
Sorry about the confusing, but this is what I mean.

Before the change:
image

After the change:
image

I still think getting the max width is some step we should do here

src/drawing/backend_impl/bitmap.rs
38: I think the reason why it can overflow is because `(alpha * 256.0).floor()` returns a 256.0, this might be possible since we have floating point issues. But if you dig into the logic at this point, we actually split the range (0.0..1.0] to 256 different alpha levels, and which means anything is greater than 255.0 / 256.0 can be treated as directly overriding, right? 
By doing so, I believe there won't be any case causing overflow anymore.

Please let me know if I was wrong.

src/chart/context.rs
nuald: Oh, I see, actually I was thinking it's a bug and never realized that it's intended :) Please let me explain - usually on the charts the right labels are aligned by the left side as the width could differ a lot. For your use-case the right alignment looks better because the digits are aligned (and for usual linear scales that probably would be better). However, for logarithmic and category labels it could be not so nice (for example, legitimate IPv4/IPv6 addresses):

boxplot

The mapping between the ticks and labels is not clearly visible. I could suggest 2 options:
1. Add an option for the alignment (now sure how the API would look like though);
2. Align based on the label source (category vs numeric etc, but again not sure is it even possible).

Unfortunately, I need the left alignment (as labels on my boxplot chart have quite different widths) and would prefer to avoid any hacks and find the proper solution.

src/drawing/backend_impl/bitmap.rs
nuald: Sorry, looks like I was not clear in the original PR message - the overflow is not happening here, but rather in the `animation` example. The fix is required to properly address invalid `alpha` values. I've added `test_draw_pixel_alphas` tests which you can use to verify the behaviour - without the fix, the bitmap code is crashing.

src/drawing/backend_impl/svg.rs
```diff
@@ -247,24 +248,36 @@ impl<'a> DrawingBackend for SVGBackend<'a> {
...
+        let ((_, x_min_y), (_, x_max_y)) =
+            font.layout_box("x").map_err(DrawingErrorKind::FontError)?;
+        let ((_, text_min_y), (_, text_max_y)) =
+            font.layout_box(text).map_err(DrawingErrorKind::FontError)?;
+        let x_height = x_max_y as f32 - x_min_y as f32;
+        let x_middle = x_min_y as f32 + x_height / 2.0;
+        let text_height = text_max_y as f32 - text_min_y as f32;
+        let text_middle = text_min_y as f32 + text_height / 2.0;
+        let scaled_height = text_height / x_height;
+        let offset = (text_middle - x_middle) / x_height;
```
38: I think the offset computing has come problem. My understanding is this offset actually makes the real bottom of the bouncing box at the `pos` location, which is exactly other parts of the code defines the locations. 
If this is the case, shouldn't it be 
```rust
offset = (text_max_y - x_max_y) / x_height;
```
As long as "x" doesn't have any part under the baseline, otherwise we may ending up with half of the size between baseline and descender. 

Correct me if I was wrong.

38: Also, I am ok with adding an interface that returns the relative descender size if you think this makes things easier. Since I believe rusttype have some way to returns the metrics of the font.

nuald: You're right that the offset calculations are slightly incorrect, but it's not as simple as just using `max_y` as it makes the vertical center incorrect with the current code. I'll update the calculations and tests with the different ascends/descends. And I doubt that any interface changes are required, `layout_box` is working good enough, it's just a matter of proper testing to finalize the calculations.

38: Yes, dig into this a little bit, I believe "dy=1em" is same as "dominant-baseline=hanging"
"dy=0.5ex" is same as "dominat-baseline=middle"  and
"dy=-0.5ex" is almost same as putting the dominant baseline to the actual bottom of the text.

How do you think about this. 

Also, I would like to say rusttype underestimates the text size a lot, so I think we should adjust the SVG's font size.

38: To explain the magic numbera 0.75, the difference between rusttype scaling and SVG scaling is SVG uses em in pixels as font size, but rusttype uses the total height of text as the size. 
So in SVG sizing model, descender isn't a part of font size, but in rusttype it also counted as the font size. 

That gives us 1em = 1 pixel is same as rusttype size 1.3. (Assuming we have descender height of 30%. And I believe the typical descender height is some value around 20% to 30%). 
That give us the factor 1 / 1.3 = 0.76 which is almost the magic number I come up with.

38: If this is the case, we actually have a pretty reliable way to define the anchor point. But the problem we have to facing is the font positioning, and my guess is this is impossible with solely SVG. 

For example, for the title spacing, we have to know the font size to decide the size of drawing area. The text size is something we have to make an reasonable guess to make it work. 

So our remaining problem could be rusttype isn't follow the SVG sizing model and that's the root for the remaining offset in the font layout. 

I actually tried render text using font-kit instead of rusttype (which I can remove a few dependency as well), it seems font-kit renders text in exact same way a browser does. Thus once we have make the change to replace rusttype, I believe we could have pretty stable SVG text layout then. 

The only thing of this plan is, changing the sizing model is definitely a breaking change which we have to wait until the next major release. 

But I believe other part of the PR seems good to me, except the label text on the right label area. 
(And I don't think this should be considered as breaking change and I believe it's ready to merge).

Sorry about the messy font handling code, as you may realized, I am just learning typography while I was creating this crate and it's a really good lesson I have learned.  

Please let me know you have any thought, and thank you so much for clean this up!

38: And update, according to wikipedia, typically, the descender height is 46% of em,which give us `dy=0.76em`
This gives us pretty much accurate result, I believe in most of the case
https://jsfiddle.net/rjmtz602/

nuald: Thank you for the research and numbers, I've updated the code and indeed it looks very good.

nuald:
I've updated the tests to verify drawing the text with the different ascends/descends. SVG and Canvas don't have the perfect pixel alignments, but nevertheless look good enough from my point of view.

Also I've discovered and fixed some minor regression issues with the category coord.

nuald:
If you don't mind, I'd prefer that you would polish the remaining items - you obviously know the context better, plus you may align your changes with your project vision. Also, I don't have the Linux desktop, only server, and actually use WSL with XServer compiled for Windows, and it has some minor issues with displaying the content (as I have the high DPI monitor that applies some scaling). So mostly all the changes are verified on Windows only (and partially on macOS), and I could definitely miss some use-cases.

38:
Of course, I verified this on Linux and MacOS. I believe the SVG rendering is better than before after the change.

This is a quite big improvement and I really appreciate for your help! 

Cheers!

38:
Finally get this merged, nice work! And I am going to ship a minor release in a week or two hopefully. Thanks for making Plotters better!

PR #99, opened 2020-01-14
Fix panic in drop

dalance:
If drop has panic, we can't handle error because panic in drop becomes abort.

### DIFF
--- a/src/drawing/backend_impl/bitmap.rs
+++ b/src/drawing/backend_impl/bitmap.rs
@@ -998,7 +998,8 @@ impl<'a, P: PixelFormat> DrawingBackend for BitMapBackend<'a, P> {
 impl<P: PixelFormat> Drop for BitMapBackend<'_, P> {
     fn drop(&mut self) {
         if !self.saved {
-            self.present().expect("Unable to save the bitmap");
+            // drop should not panic, so we ignore a failed present
+            let _ = self.present();
         }
     }
 }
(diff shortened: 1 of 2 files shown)


38:
Hi @dalance , thanks for the PR. The only thing in my mind is I think we should do something on failure, instead of ignoring it. Since it's weird if the program exit normally, but there's no output. 
So I feel it's better to have an error message in this case.

dalance:
I want to handle the error like below:

```
// chart is plotters::chart::ChartContext
chart
    .plotting_area()
    .present()
    .context(format!("failed to write SVG: {:?}", path))?;
```

By this code, I expect the following error message.

```
failed to write SVG: "/a.svg"
  Caused by: backend error: Drawing backend error: Permission denied (os error 13)
```

But the actual message becomes below:

```
thread 'main' panicked at 'Unable to save the SVG image: DrawingError(Os { code: 13, kind: PermissionDenied, message: "Permission denied" })', src/libcore/result.rs:1165:5
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace.
```

I think this is not user-friendly, but I can't customize the message because panic in drop.

38:
Panic in drop is definitely not a good behavior. I just feel uncomfortable with ignoring error either. 

But yes, you are right printing an error message might not be friendly to everyone and probably mess up the output. Just looked how standard library handles error during dropping RAII, it seems they also ignores any error.  

I probably OK with ignoring it now. Thanks for the explain.

PR #119, opened 2020-03-16
set svg viewBox

elwerene:
Is needed for scalable display in browser


38:
Awesome, thanks for the contribution!

PR #124, opened 2020-03-27
Fix panic in console.rs example

Eosis:
Hi all,

I had a panic when running this example while checking out the library: The example PNG was being written to a non-existent folder. I've altered this and put a print in to say where the PNG is printed.

Let me know if you have any feedback or if I've misunderstood something. 😄

### DIFF
--- a/examples/console.rs
+++ b/examples/console.rs
@@ -187,9 +187,10 @@ where
 
 fn main() -> Result<(), Box<dyn Error>> {
     draw_chart(TextDrawingBackend(vec![PixelState::Empty; 5000]).into_drawing_area())?;
-    let b = BitMapBackend::new("plotters-doc-data/console-example.png", (1024, 768))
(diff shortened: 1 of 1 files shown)


38:
Hi, 

Thanks for the PR. I understand this is confusing when people just run the example code and failed. But the plotters-example directory in the repo and this is used as a validation when layout code was changed. 

For details, when we want to change the layout code, we can run all the examples it will automatically produce new output to the `plotters-doc-data` directory, which is tracking by another module. By doing so, we can avoid some unexpected layout change. 

But I know this is confusing and some people complains issues regarding example not working. So what I would like to do is add some instruction on how to run the examples rather than change the outputting directory.

Closing the PR and create an issue to track the documentation stuff #125  , if you have any idea on this, please let me know.

Thanks again for the contribution
Cheers

PR #139, opened 2020-05-04
Format Cargo.toml

TatriX:
Replace tabs with spaces and reformat the file to get rid of the super long lines.


38:
Neat! Thanks so much!

PR #147, opened 2020-05-27
adjusts html canvas backend size using dpr

Marius-Mueller:
Using get_bounding_client_rect to get the html canvas size causes the graph to be incorrectly sized if certain ccs styling is applied to the canvas.  
For example, if you use css to scale a chart, get_bounding_client_rect will return the size to which the canvas will be scaled, not the actual size of the canvas. This causes the plot to be drawn with the wrong dimensions if the css scaling isn't set to the precise values which compensate for the dpr, either clipping off the canvas if it is scaled up, or only taking up part of the canvas if it is scaled down.

To address this issue, I use the device pixel ratio to properly adjust the size of the canvas for high DPI screens.

This is essential for resizing graphs without re-rendering them.


38:
Thanks awesome, thanks so much!

PR #149, opened 2020-05-31
Update to rusttype 0.9

jrmuizel:
This adds support for fonts with CFF glyph outlines.

### DIFF
--- a/src/style/font/ttf.rs
+++ b/src/style/font/ttf.rs
@@ -2,10 +2,10 @@ use std::borrow::{Borrow, Cow};
 use std::collections::HashMap;
 use std::i32;
 use std::io::Read;
-use std::sync::{Arc, RwLock};
+use std::sync::RwLock;
 
 use lazy_static::lazy_static;
-use rusttype::{point, Error, Font, FontCollection, Scale, SharedBytes};
+use rusttype::{point, Font, Scale};
 
 use font_kit::family_name::FamilyName;
 use font_kit::handle::Handle;
@@ -20,7 +20,7 @@ type FontResult<T> = Result<T, FontError>;
 pub enum FontError {
     LockError,
     NoSuchFont(String, String),
-    FontLoadError(Arc<Error>),
+    FontLoadError,
 }
 
 impl std::fmt::Display for FontError {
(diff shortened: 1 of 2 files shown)


38:
Thanks for contributing to Plotters. The most part of the change seems good to me, but updating to rusttype 0.9 seems break the compilation in lower rustc version (Currently we are maintaining MSRV 1.36.0 and for v0.2.x it could causing some downstream project fail to compile). So I would like to keep current rusttype version back in plotters 0.2.x  

But I would like to revisit this PR when we am ready to release v0.3 (which I am currently working on and hopefully this is going to happen soon). At that time we can make breaking change include bumping up the MSRV. 

Thanks again for the contribution!

src/style/font/ttf.rs
```diff
@@ -100,10 +100,8 @@ fn load_font_data(face: FontFamily, style: FontStyle) -> FontResult<Font<'static
         };
         // TODO: font-kit actually have rasterizer, so consider remove dependency for rusttype as
         // well
-        let result = FontCollection::from_bytes(Into::<SharedBytes>::into(data))
-            .map_err(|err| FontError::FontLoadError(Arc::new(err)))?
-            .font_at(id.max(0) as usize)
-            .map_err(|err| FontError::FontLoadError(Arc::new(err)));
+        let result = Font::try_from_vec_and_index(data, id.max(0))
```
38, line 103: Seems rusttype 0.9 doesn't have font loading error information any more, just wondering is there any way to distinguish the different kind of loading error. For example, "UnrecognizedFormat" vs "CollectionIndexOutOfBounds" https://docs.rs/rusttype/0.8.3/rusttype/enum.Error.html

RazrFalcon: `rusttype` uses [ttf-parser](https://github.com/RazrFalcon/ttf-parser) now, which doesn't report any errors. Mainly because there are not much to tell. The API is designed around the idea that each method can produce only one kind of error. Therefore we are using `Option`.

For example, we can return an error saying that the CFF table has an outdated type1 format, but it doesn't really tell anything to a caller since you have to know all the TrueType nuances to reason about this.

38:
Hi there, github seems doesn't recognize the merge https://github.com/38/plotters/commit/687db8. Closing the PR

PR #156, opened 2020-07-03
HTML5 backend: line width and drawing area size fixes

AdrienChampion:
Fixes the handling of `stroke_width` from [`ShapeStyle`](https://docs.rs/plotters/0.2.15/plotters/style/struct.ShapeStyle.html) when drawing lines in the canvas backend.
Most line-like (line, path, circle, rect) drawing was setting the stroke style, but not the line width (using [`set_line_width`](https://rustwasm.github.io/wasm-bindgen/api/web_sys/struct.CanvasRenderingContext2d.html#method.set_line_width)). So in these cases, the line width actually used was the default or the value provided in the latest call to the rare functions that actually called `set_line_width`.

- introduced function `set_line_style` for `CanvasBackend` which sets the stroke style and the line width in the context
- now calling that function when needed, *i.e.* in
    - `draw_line`, `draw_path`, and
    - `draw_rect`, `draw_circle` (when not filling)

Fixes device-pixel-ratio/canvas-size problems introduced in https://github.com/38/plotters/commit/1b4e11a11cddded2cbe348826d7da010181640e5#diff-31f17ae10d710e88872f9010d97ec041. I do not understand plotters enough to give a clear explanation of the problem, but my *vague* understanding that is that the current version

- is using the wrong width/height (the ones directly from the canvas instead of using the bounding rectangle)
- is dividing by the DPR instead of multiplying by it

I might be wrong, and/or my fix might not make sense logically. Do let me know if that's the case. What I can say is that the current version looks terrible (very pixelated / not scaled properly) on high (and possibly low) DPI *in my experiments*, which is not the case of the version I am proposing here.

### DIFF
--- a/src/drawing/backend_impl/canvas.rs
+++ b/src/drawing/backend_impl/canvas.rs
@@ -84,9 +84,10 @@ impl DrawingBackend for CanvasBackend {
         let window = window().unwrap();
         let mut dpr = window.device_pixel_ratio();
         dpr = if dpr == 0.0 { 1.0 } else { dpr };
+        let rect = self.canvas.get_bounding_client_rect();
         (
-            (self.canvas.width() as f64 / dpr) as u32,
-            (self.canvas.height() as f64 / dpr) as u32,
+            (rect.width() as f64 * dpr) as u32,
+            (rect.height() as f64 * dpr) as u32,
         )
     }
 
--- a/src/drawing/backend_impl/canvas.rs
+++ b/src/drawing/backend_impl/canvas.rs
@@ -61,6 +61,13 @@ impl CanvasBackend {
     pub fn with_canvas_object(canvas: HtmlCanvasElement) -> Option<Self> {
         Self::init_backend(canvas)
     }
+
+    /// Sets the stroke style and line width in the underlying context.
+    fn set_line_style(&mut self, style: &impl BackendStyle) {
+        self.context
+            .set_stroke_style(&make_canvas_color(style.as_color()));
+        self.context.set_line_width(style.stroke_width() as f64);
+    }
 }
 
 fn make_canvas_color(color: RGBAColor) -> JsValue {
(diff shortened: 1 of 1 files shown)


src/drawing/backend_impl/canvas.rs
```diff
@@ -77,8 +84,11 @@ impl DrawingBackend for CanvasBackend {
         let window = window().unwrap();
         let mut dpr = window.device_pixel_ratio();
         dpr = if dpr == 0.0 { 1.0 } else { dpr };
-        ((self.canvas.width() as f64 / dpr) as u32,
-         (self.canvas.height() as f64 / dpr) as u32)
+        let rect = self.canvas.get_bounding_client_rect();
```
38, line 87: Thanks for point this out, but I believe this is an actually wrong fix to another issue. 

Basically, the invariant here we want to maintain is in: If plotters want to draw a pixel at (width, height) it should present in the right bottom, that is the right positioning. As long as the CanvasBackend dynamically get the logical size of the canvas. So this part shouldn't related to DPR at all. 

Now I just realized this is a really complicated story:

I previously merged a PR which does wrong scale adjustment on DPR (in the WASM example's JS code). I found the similar method is described [here](https://www.html5rocks.com/en/tutorials/canvas/hidpi/) but the whole point is the the blog post assume the drawing code hard coded the canvas size, thus once we adjust the canvas logic size, we should then scale it, so that we don't need to modify the code. But Plotters does a completely different thing, the backend check the current logic size each time and all other part of the crate uses this. But it looks Ok on normal DPI screen.


After that somebody else realize that on a highDPI screen it has a problem, which comes up with this "fix". Which divide the DPR so that this can cancel the previously wrong scale out. But this is not really a fix.

Thanks for point this out and I am going to clean up the mess soon. If you don't mind could you please change this to 
```
(self.canvas.width(), self.canvas.height())
```
This is what it should be

AdrienChampion:
Sorry it took me so long to get back to you... I see you merged the PR, do you need me to do anything at this point?

38:
Yep, I just fixed the previously introduced bug based on your branch and merged it.  Thanks a lot!

PR #171, opened 2020-07-27
Add cargo fmt to travis build config

Atul9:


38:
Hi sorry for getting back so late. But I am a little bit hesitate about adding this to CI - since formatting is something that is softly enforced, and this may be a little bit bothering to people.

And in addition, rustfmt seems not too stable. Thus I would like not to have that in CI.

Let me know if you have any thoughts. Thanks so much!

PR #176, opened 2020-08-15
Feature: new `Quartiles` from precalculated values

evanrichter:
I recently had a dataset that took hours to generate, so I wanted a workflow like:

1. generate dataset once
2. save quartile data to csv
3. plot csv with plotters

This PR allowed me to provide pre-calculated values for the quartile range as desired

This implementation simply panics if the values aren't in a sensible order

### DIFF
--- a/src/data/quartiles.rs
+++ b/src/data/quartiles.rs
@@ -62,6 +62,33 @@ impl Quartiles {
         }
     }
 
+    /// Create a new quartiles struct with pre-calculated values.
+    ///
+    /// - `s`: The array of the original values
+    /// - **returns** The newly created quartiles
+    ///
+    /// ```rust
+    /// use plotters::prelude::*;
+    ///
+    /// let quartiles = Quartiles::new_from_values(&[2, 24, 31, 39, 45]);
+    /// assert_eq!(quartiles.values(), [2.0, 24.0, 31.0, 39.0, 45.0]);
+    /// ```
+    pub fn new_from_values<T: Into<f64> + Copy + PartialOrd>(s: &[T; 5]) -> Self {
+        let s = s.to_owned();
+        assert!(s[0] <= s[1]);
+        assert!(s[1] <= s[2]);
+        assert!(s[2] <= s[3]);
+        assert!(s[3] <= s[4]);
+
+        Self {
+            lower_fence: s[0].into(),
+            lower: s[1].into(),
+            median: s[2].into(),
+            upper: s[3].into(),
+            upper_fence: s[4].into(),
+        }
+    }
+
     /// Get the quartiles values.
     ///
     /// - **returns** The array [lower fence, lower quartile, median, upper quartile, upper fence]


src/data/quartiles.rs
```diff
@@ -62,6 +62,33 @@ impl Quartiles {
...
+    ///
+    /// ```rust
+    /// use plotters::prelude::*;
+    ///
+    /// let quartiles = Quartiles::new_from_values(&[2, 24, 31, 39, 45]);
+    /// assert_eq!(quartiles.values(), [2.0, 24.0, 31.0, 39.0, 45.0]);
+    /// ```
+    pub fn new_from_values<T: Into<f64> + Copy + PartialOrd>(s: &[T; 5]) -> Self {
+        let s = s.to_owned();
+        assert!(s[0] <= s[1]);
```
38, line 78: In general, we don't want to panic even if things are going wrong, since we are a library, we want to avoid this kind of panic.

src/drawing/backend_impl/piston.rs
```diff
@@ -34,6 +34,14 @@ fn make_point_pair(a: BackendCoord, b: BackendCoord, scale: f64) -> [f64; 4] {
     ]
 }
 
+fn make_circle(center: BackendCoord, radius: u32, scale: f64) -> [f64; 4] {
```
38, line 37: To me this isn't related to the PR, right?

38 (changes requested):
Hi sorry for getting back to you so late, I looked at the code, it's goods good to me mostly. But I don't think we should panic at this point anyway, as long as we are a library we should avoid any unexpected inputs and returns error in most of the case. 

 Thanks again for the help!

src/drawing/backend_impl/piston.rs
evanrichter: I think @pelekhay added this commit

src/data/quartiles.rs
evanrichter: We could either:

1. collect the array into a Vec then sort
2. change the return type to Result<Self, SomeErrorType>
3. leave it

Personally I would choose 3 (perhaps with an assertion string to explain what happened). If I calculate or receive quartile values and they are not in order, I want to know very directly that something is wrong.

1 is my next choice, but it could silently smooth over some logic error the user made earlier.

2 has both the previous benefits, but a big fallback since users now have to handle an error variant for what should be a very simple operation.

What do you think?

38: I don't think we should panic here, since it doesn't deserve that - more importantly, panicking in a library is more like an "unrecoverable error". This is definitely not "an unrecoverable error". In today's Plotters, there are places may panic unexpectedly, but we are constantly working to get rid of them.  

In principle, we don't allow any panic unless there's no way to be correctly handled or we believe panicking is impossible to happen.

I am fine with both 1 and 2, but I am personally prefer 1. Returning result is still an overkill (And panicking in lib is even bad than that). 

Also if you already copied the data, sorting a 5 elements slice is very lightweight.

src/drawing/backend_impl/piston.rs
38: So would you mind rebase to the master branch - since you are actually based on the 0.2 branch

evanrichter: ah my bad, I will rebase

src/data/quartiles.rs
evanrichter: Ok sounds good, I'll implement option 1 then!

PR #183, opened 2020-09-09
Remove references to `build_ranged` in examples

chrisduerr:
Since the function is deprecated, it shouldn't be recommended in
examples. All instances of `build_ranged` have been replaced with the
new `build_cartesian_2d`.


38:
Thanks so much for the contribution!

PR #185, opened 2020-09-10
Allow passing colors by value and reference

chrisduerr:
This fixes a bit of a usability issue with the API of plotters where
certain functions only take references to colors, even if an owned value
is easily available. Especially if the same builder pattern takes
multiple colors, it's extremely annoying if the color cannot be
initialized once and then copied around.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,7 +15,7 @@ exclude = ["doc-template/*"]
 [dependencies]
 num-traits = "0.2.11"
 chrono = { version = "0.4.11", optional = true }
-plotters-backend = "^0.3.*"
+plotters-backend = { git = "https://github.com/chrisduerr/plotters-backend", rev = "529f08f3aa5ede8e71950c1a5d3be4683330f549" }
 plotters-svg = {version = "^0.3.*", optional = true}
 
 [dependencies.plotters-bitmap]
--- a/src/style/color.rs
+++ b/src/style/color.rs
@@ -50,9 +50,11 @@ pub trait Color: BackendStyle {
     }
 }
 
+impl<T: Color> Color for &'_ T {}
+
 /// The RGBA representation of the color, Plotters use RGBA as the internal representation
 /// of color
-#[derive(Clone, PartialEq, Debug)]
+#[derive(Copy, Clone, PartialEq, Debug)]
 pub struct RGBAColor(pub(crate) u8, pub(crate) u8, pub(crate) u8, pub(crate) f64);
 
 impl BackendStyle for RGBAColor {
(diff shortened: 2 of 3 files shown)


chrisduerr (commented):
This requires the following patch to plotters-backend:

```diff
diff --git a/src/style.rs b/src/style.rs
index 28af116..4191831 100644
--- a/src/style.rs
+++ b/src/style.rs
@@ -30,3 +30,9 @@ impl BackendStyle for BackendColor {
         *self
     }
 }
+
+impl<T: BackendStyle> BackendStyle for &'_ T {
+    fn color(&self) -> BackendColor {
+        todo!();
+    }
+}
```

I did not find this hosted anywhere, so I cannot send a PR.

38:
Hi, thanks for the PR. And it's nice to improve the usability. 

For better understanding the issue this PR addressed, could you please give an example? - Just wondering if we can fix it without touching the plotters-backend crate.

Thank you so much!

chrisduerr:
I don't have the code around anymore, but it's basically like this (adapted from the example):

```rust
    let color = RED;
    chart
        .draw_series(LineSeries::new(
            (-50..=50).map(|x| x as f32 / 50.0).map(|x| (x, x * x)),
            &color,
        ))?
        .label("y = x^2")
        .legend(|(x, y)| PathElement::new(vec![(x, y), (x + 20, y)], &color));
```

When compiling that, it shouldn't work since you're using the same reference twice. But it obviously makes a ton of sense to do that here, since you basically always want to use the same color for both. I had an array of colors defined and indexed it with a remainder to dynamically pick out the colors, so it was longer than just something short like "RED". But I think that's not uncommon.

38:
Hi there, I think one of the reason why we need change `plotters-backend` is actually how the `Color` trait is defined. 

`BackendColor` actually used as the unified color representation used by all the Plotters backend code, thus the backend do not need to consider the complexity of the color representation. 

`BackendStyle` is a trait that primarily used for responsibilities, since once we want to pass more information to the backends, adding BackendStyle trait method will avoid breaking change to both core and backends code. 

In this particular `Color` trait, the only reason for it impls `BackendStyle` is to call the `color` method which convert any color to `BackendColor`.  So that I believe there's no real reason for any type T to be a `BackendStyle` but having a method doing color conversion. 

However, if we loose the requirement for `Color: BackendStyle`, we are introducing breaking change to anyone who use the low-level drawing APIs. But I guess this is ok, since I don't believe there's a lot people use low-level drawing API in their code, plus we actually can impl BackendStyle for the most commonly used colors for example predefined colors in the macro. 

How do you think about this? I believe if we move this way, we can simplify the Color trait a little bit

38:
Hi, just tried to made some change on color trait and I believe this is still work for you, but it doesn't requires to change the backend crate. 

To me, I want to 1) make the backend crate stable 2) decouple backend types from the core types. 

But I think this makes a bigger change, please let me know how do you think about this

Thanks!

chrisduerr:
I personally do not care much tbh. I've just played around with the crate a bit and immediately ran in a few issues and thought I'd contribute back to improve things.

If you have some concrete plans on how you might want to resolve the same issue, I'd recommend you go aheand and implement that yourself. Just close this PR if you have found a different solution.

In general I'd just recommend to create an API that doesn't restrict the user to passing colors in as reference, when these can be easily made `Copy`.

38:
Sure, so I just merged this PR with the change I've made. Thanks!

facorread:
Close #185

PR #197, opened 2020-10-26
improve the bivariate normal pdf 3d plot example

kineticz:
Used formula from wolfram mathworld to define the pdf function.


kindlychung:
3d-plot2

el-hult:
What is the problem you are trying to solve with this pull request? It seems to me the code just got more complicated with no clear benefit.

philocalyst:
Agree with @el-hult , I would motion for this to close @AaronErhardt

PR #201, opened 2020-11-06
Removed broken link and picture

DesmondWillowbrook:
Removed broken link which is supposed to point to demonstration of Piston backend, which has been removed.


jayvdb:
This has been solved on `master`, and the PR can be closed

PR #215, opened 2021-01-07
Fix margins in Horizontal Histograms

adriandelgado:
This fixes a bug where the Iterator implementation for Horizontal Histogram was setting the margins in the same way as the Vertical Histogram.


adriandelgado:
Friendly ping. It's just a small bug fix, probably a result of copy-pasting the same code.

facorread (approved):
Thanks for your contribution!

PR #220, opened 2021-01-19
Exclude target_os=wasi from browser

### LINKED ISSUE #219: Support for wasm32-wasi target
I ran into an issue while trying to compile plotters to WebAssembly.

Currently **any** `wasm32-*-*` target (e.g. `wasm32-unknown-unknown`, `wasm32-wasi, `wasm32-...`) is considered to be running inside the browsers and uses `wasm-bindgen` as a dependency.

But this is not always the case. In my case I'm trying to run it on the backend inside an application compiled to `wasm32-wasi` on [Wasmtime](https://github.com/bytecodealliance/wasmtime) and [Wasmer](https://github.com/wasmerio/wasmer). The issue is that `wasm-bindgen` doesn't work on this platforms because they are not browsers and the required wasm-bindgen imports are not satisfied.

Would plotters still work if we just treated `wasm32-wasi` as a non-wasm backend?

bkolobara:
Closes #219

There are [many new](https://wasmer.io/) [WebAssembly](https://wasmtime.dev/) runtimes intended to be run outside of the browsers. This pull request makes plotters compatible with them.

The main issue is that `wasm-bindgen` doesn't make sense when the target os is specified as `wasi` and there is no JavaScript environment. Wasi already exposes low level APIs that can be used instead.

A nice side-effect of this pull request is that you can now choose to compile with `target=wasm32-wasi` and still run your .wasm file in the browser, but provide a WASI polyfill instead of `wasm-bindgen`.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -23,22 +23,22 @@ version = "^0.3.*"
 optional = true
 default_features = false
 
-[target.'cfg(not(target_arch = "wasm32"))'.dependencies]
+[target.'cfg(not(all(target_arch = "wasm32", not(target_os = "wasi"))))'.dependencies]
 ttf-parser = { version = "0.8.2", optional = true }
 lazy_static = { version = "1.4.0", optional = true }
 pathfinder_geometry = { version = "0.5.1", optional = true }
 font-kit = { version = "0.7.0", optional = true }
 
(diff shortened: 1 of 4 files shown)


gilescope:
Cool as. The more stuff that compiles to wasm the better.

PR #226, opened 2021-01-31
Fillable candlestick.

attila-kun:
This PR allows filling the inside area of candlesticks:

stock


inconspicuous99:
what happened to this one?

facorread:
@inconspicuous99 This PR failed both checks; the project has implemented new checks since.

Would you like to propose an updated PR and some tests? The project would appreciate it.

38:
Seems we already have merged a PR that fix the bug that candlestick element discard the filled field from the style. 

See https://github.com/plotters-rs/plotters/commit/b41b5a7a2a53745a30c51d6fe4024f7b2df8e7f9 

Plus this change just make the Candlestick::new doesn't compatible with previous version, thus it's a breaking change. So I am prefer the fix I linked above. 

Thanks for your contribution and please don't mind I close the PR.

PR #237, opened 2021-03-11
Introduction of Boxplots with outliers

tobiasvandriessel:
I added a Boxplot variant with outliers, but I don't know whether it's built the way you'd have liked, as I can imagine you would have liked to have it as a variant of a generic type. However, I decided to build something quick and dirty and figured you might as well think it's okay. I'm open to any suggestions, but I'm still a Rust beginner, so I'm not sure I can do exactly what you'd like.

Cheers.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "plotters"
-version = "0.3.0"
+version = "0.3.1"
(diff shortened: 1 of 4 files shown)


wangjiawen2013:
Hi, do you resolve the conflicts ?

PR #249, opened 2021-05-06
update font-kit

zyansheep:
Conflicting with other library using different version of font-kit.


38:
Hi there, thanks for the PR! 

But we are not accepting any dependency updating PRs, as we use dependabot handle it automatically.

PR #254, opened 2021-05-22
Enhance documentation of predefined colors

siefkenj:
The predefined colors are now documented with their RGB(A) values and a box colored with the corresponding color.


38:
That's awesome! thanks for contribution!

PR #256, opened 2021-05-23
Add full color palette with example

siefkenj:
Adds many more pre-defined colors accessible via `full_palette::<color name>` as well as an example demonstrating all of their use. The colors are defined by the Material UI standard to work well together. I've included a rendering of the example.

full_palette

Maybe at some point these colors can replace the current default colors. I think it would make for more readable charts with less effort.


siefkenj:
Based on the outcome of the coverage report, I assume that the examples are not included?

38:
This looks awesome, thanks for working on this!

> Based on the outcome of the coverage report, I assume that the examples are not included?

There's some problem with the coverage tool, but I think that's fine.

38:
Just merged the PR, this looks really nice! Thanks for contributing!

Issue #258, opened 2021-05-25
New chart layout algorithm


siefkenj:
I am working on a new chart layout algorithm that may be slightly controversial, so I wanted to get some feedback before I make a PR.

My goal is to make charts appear nicer by default. For example, to have margins/padding automatically computed based on font size and to prevent axis labels from printing off the edge as in 

![graph](https://camo.githubusercontent.com/28a0d2b476dc639598edc6f06291f394d4abd678b86b2b25218faf35c9923043/68747470733a2f2f706c6f74746572732d72732e6769746875622e696f2f706c6f74746572732d646f632d646174612f302e706e67)

The idea I had was to use the [stretch](https://lib.rs/crates/stretch) library, which performs flex-box calculations in Rust (and is WASM compatible and supposedly very efficient). Then, a chart could be laid out as a series of nested flex boxes, and a layout could be computed when needed, instead of immediately as in the current `ChartBuilder` api.

If a layout is set up and then computed, this means it can have default font sizes, etc, which can be overwritten, instead of everything having to be set upfront. Size calculations can then be done dynamically.

So far I have decomposed a chart into regions as follows:

![layout](https://user-images.githubusercontent.com/839083/119568303-5931e000-bd7b-11eb-8b56-0963d5c18baf.png)

Any region can have a size set to zero and it will disappear from the layout. I believe the above decomposition will be able to handle all current use cases. Please correct me if I'm wrong!

Is this something that would make it into standard `plotters`? Behind a config flag maybe?

38:
Hi, thanks for working on this. In general this sounds great to me, and it would be awesome if Plotters can do some automatic adjustment by default.  There are few things coming out from my mind at this time:

1. We probably can have this behind some feature switch until we reach next major release - Basically we don't want any existing downstream code see a lot of layout changing by default (some of them may be disaster especially for criterion) 
2. There are one tweak may be interesting: Plotters allows the label area float over the chart area by passing a negative label_area_size. Also the negative tick mark size make them drawn inward. 
3. Sometimes we want to have the layout stable for realtime plotting - I mean for each frame, we basically want to have exactly same layout for each frame. By doing so, we can only updating the chart_area without redraw rest part of the image. 

Also, I am curious is the new layout algorithm able to automatically layout multi-panel charts?

BTW, as the scratch crate doesn't introduces tons of dependencies, I would like to make it default in next major release if everything goes well.

Really appreciate for the contribution!

siefkenj:
[...]

Issue #189, opened 2020-09-18
[BUG] No warning for when folder doesn't exist


JacobHaig:
**Describe the bug**
I was following the quickstart guide example on https://docs.rs/plotters/0.3.0/plotters/ and I noticed that the image was not generating when the folder doesn't exist. There is no warning or error. It seems like it is intended that if the folder doesn't exist, the folder would be created.

**To Reproduce**
let root = BitMapBackend::new("folder-that-doesnt-exist/0.png", (640, 480)).into_drawing_area();
needs to be changed to 

let root = BitMapBackend::new("folder-that-exists/0.png", (640, 480)).into_drawing_area();
or
let root = BitMapBackend::new("0.png", (640, 480)).into_drawing_area();

**Version Information**
plotters = "0.3" from Crates.io

38:
Hi there, thanks for trying Plotters!

This is actually a know behavior, but you are right, it's confusing. Originally, it panic during drop - bitmap only encode the image file when the bitmap backend is being dropped - But this causing troubles and we have a fix for this #99 

Note the code has been moved to https://github.com/plotters-rs/plotters-bitmap/blob/3891badaf5916cd1ffbbd55dd21239e671a4b0b8/src/bitmap.rs#L718 since 0.3

But I realized current behavior is very confusing, since it silently drop any IO error - But in fact the rust std did the same thing. But I believe the solution would when the bitmap backend is created we check if the file is writable first. 

However, this introduce a breaking change, since current API doesn't returns a result. 

So I guess we may want to add a new API let's say `create` to plotters-bitmap backend which check if the file writable and return error when it's impossible to write the target file. And finally we can deprecate what we currently have, at least we can mark the `BitMapBackend::new` deprecated once we provided a new method.

Let me know if you have any ideas!

Thanks,
Hao

ijackson:
Hi.  Thanks for this library.  I just encountered this situation. 

I found the behaviour quite disturbing and not what I expected from a Rust program: it's normally quite difficult to accidentally write a Rust program which ignores file write errors.  Having read the issue et . I see that it's not entirely trivial to fix. IMO the root cause is the use of `drop` to do serious amounts of work, including fallible work.  That doesn't work very well, as we see.

For now, I think a short term answer is to have everyone who uses `BitMapBackend::new()` also call `.present()?`.  I'll submit MRs shortly to do that in the examples, and add a note to the documentation.

[...]

PR #259, opened 2021-05-26
Advanced Flexbox Layout

siefkenj:
Here's some initial work towards ideas pointed out in #258 

Most of the work so far is in `nodes.rs`, which builds the layout and interfaces with the `stretch` library. I am now working on `ChartLayout` which should be similar to `ChartBuilder` but with additional flexibility. However, I could use some pointers...

Something with how I am doing error handling in `ChartLayout` isn't right. In `examples/layout.rs`, if I use `?` to unwrap the errors, I get a compile error:
```
returns a value referencing data owned by the current function
```
which I assume is because the error is somehow related to the `DrawingArea` to which I hold a mutable reference. However, I tried to mirror your code closely, so I'm not sure what's wrong. (If I use `.unwrap()` instead of `?` it works.) Any ideas?


siefkenj:
I think I figured it out...I switch to concrete error types instead of `dyn std::error::Error` and now it works. I don't know why the `dyn Error` was causing issues, though...

siefkenj:
@38 The bulk of the API should be implemented now. Please let me know what you think.

One question I had is about the `LayoutError` enum. I modified it so that it could encapsulate errors for `ChartLayout`. It doesn't look like `LayoutError` is used anywhere in `plotters` or any of the plotters backends. Does anyone use this? Is it a breaking change?

So far I have implemented labels and extents, but not yet the passdown to the sub-region (`chart_area`) where plotting actually takes place. I plan to do something similar to `ChartBuilder` and just pass back a drawing area restricted to that location. The only thing I really need to work out is how to figure out if a label will spill into the margin so that I can automatically adjust the margin size.

Here's an example of some ouput of the current code. Sorry for the huge codedump!

layout2

```rust
use plotters::prelude::*;

const OUT_FILE_NAME: &'static str = "plotters-doc-data/layout2.png";

    const W: u32 = 600;
    const H: u32 = 400;
    let root = BitMapBackend::new(OUT_FILE_NAME, (W, H)).into_drawing_area();
    root.fill(&full_palette::WHITE)?;

    let mut chart = ChartLayout::new(&root);
    chart
        .set_chart_title_text("Chart Title")?
        .set_top_label_text("A label at the top")?
        .set_chart_title_style(("serif", 60.).into_font().with_color(&RED))?
        .set_left_label_text("Left label")?
        .set_right_label_text("Right label")?
        .set_bottom_label_text("Bottom label")?
        .set_bottom_label_margin(10.)?
        .set_top_label_margin(10.)?
        .draw()?;

    let extent = chart.get_chart_area_extent()?;
    root.draw(&Rectangle::new(extent.into_array_of_tuples(), &BLUE))?;
    let extent = chart.get_chart_title_extent()?;
    root.draw(&Rectangle::new(extent.into_array_of_tuples(), &BLUE))?;
    //dbg!();

    Ok(())
}

```

siefkenj:
This should be ready now. I currently just pass a sub-drawing area where the chart area is to allow for graphing. I tried to figure out how to get tick locations so I could compute their size and automatically resize the chart, but it was just too hard and I couldn't find out the types.

For now we get an output that looks like

layout2

as a result of `examples/layout.rs`. I think maybe the automatic sizing of the tick labels should come in a separate PR, since this one s already getting pretty long.

Please let me know what you think!

src/chart/layout/nodes.rs
```diff
@@ -0,0 +1,816 @@
...
+    geometry::Size,
+    node::{MeasureFunc, Node, Stretch},
+    number::Number,
+    style::*,
+};
+
+/// Trait to constrain a type to a standard numeric type
+/// to prevent ambiguity in trait definitions.
+/// Idea from: https://stackoverflow.com/questions/39159226/conflicting-implementations-of-trait-in-rust
+pub trait Numeric {}
```
38, line 19: I am not sure if this is necessary. Also just curious if this is needed, can we just use https://docs.rs/num/0.4.0/num/trait.Num.html ?

src/chart/layout/mod.rs
```diff
@@ -0,0 +1,400 @@
+use crate::{coord::Shift, style::IntoTextStyle};
```
38, line 1: Just saw a few warning there, we need to make sure our code doesn't have warning

src/drawing/area.rs
```diff
@@ -130,6 +130,14 @@ impl<DB: DrawingBackend, CT: CoordTranslate + Clone> Clone for DrawingArea<DB, C
     }
 }
 
+/// Sub error type specific to layout errors
+#[derive(Debug)]
+pub enum LayoutError {
+    /// Layout error coming from the Stretch library
+    StretchError(stretch::Error),
```
38, line 137: Maybe a `#[cfg(feature = "stretch")]`. I think it's better to having this feature optional for now

src/style/palette.rs
```diff
@@ -1,4 +1,11 @@
-use super::color::PaletteColor;
+use super::{color::PaletteColor, full_palette};
```
38, line 1: Note that full_palette is optional, so a conditional complication is required at this point

examples/layout.rs
```diff
@@ -0,0 +1,58 @@
...
+    const W: u32 = 600;
+    const H: u32 = 400;
+    let root = BitMapBackend::new(OUT_FILE_NAME, (W, H)).into_drawing_area();
+    root.fill(&full_palette::WHITE)?;
+
+    let mut chart = ChartLayout::new(&root);
+    chart
+        .set_chart_title_text("Chart Title")?
+        .set_chart_title_style(("serif", 60.).into_font().with_color(&RED))?
+        .set_left_label_text("Ratio of Sides")?
```
38, line 15: Also, this seems odd as we also have https://docs.rs/plotters/0.3.1/plotters/chart/struct.MeshStyle.html#method.x_desc which have the similar effect.

38:
Hi Jason,

Thanks for the PR, I am so excited to try this out. The general idea looks really interesting and I feel this actually makes plotters easier to use. 

In general, I think the API workflow is a little bit odd. Currently, if my understand is correct, we need to build a `ChartLayout` and then pull out the chart_area from the layout then configure the chart context.  But could we just build a chart context on top of `ChartLayout`?

I am pretty excited about what the current PR have, but I probably have some idea to tweak the workflow later. It looks good to me with the API if we can put it behind a feature switch. 

Also I realized there are few warning introduced by the PR, we need to fix them.  

Also, please don't mind I was not able to see this during the last weekend. 

Thanks for this awesome PR!
Cheers!

examples/layout.rs
```diff
@@ -0,0 +1,58 @@
...
+
+    cc.draw_series(LineSeries::new(
+        x_axis.values().map(|x| (x, x.cos())),
+        &BLUE,
+    ))?
+    .label("Cosine")
+    .legend(|(x, y)| PathElement::new(vec![(x, y), (x + 20, y)], &BLUE));
+
+    cc.configure_series_labels().border_style(&BLACK).draw()?;
+
```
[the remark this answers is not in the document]
38: See issue #189

src/chart/layout/mod.rs
```diff
@@ -0,0 +1,400 @@
...
+    pub fn new(root: &'a DrawingArea<DB, Shift>) -> Self {
+        Self {
+            root_area: root,
+            chart_title: Label {
+                text: None,
+                style: TextStyle::from(("serif", 40.0).into_font()),
+            },
+            top_label: Label {
+                text: None,
+                style: TextStyle::from(("serif", 25.0).into_font()),
```
38, line 166: Actually, you can just use `("serif", 25.0).into()`.

38: Also, if we want to optimize the default appearance, I think we should use relative sizing.

examples/layout.rs
siefkenj: I guess this should be a broader discussion about how charts should be configured and what belongs to the chart vs. the coordinate system.

The way plotters seems to do it now is to attach labels to the coordinate system and then adjust whether the axes are printed to the left right/etc. Since `ChartLayout` doesn't handle coordinate systems right now, I just went an API that labels things based on position.

I think that maybe the best solution would be to have both. If you set via the coordinate system, it would automatically call the `set_XXX_label_text`. However if a person wanted to set the text without a coordinate system or to override it, they would be free to.

What do you think?

Cargo.toml
```diff
@@ -16,6 +16,8 @@ exclude = ["doc-template/*"]
 num-traits = "0.2.14"
 chrono = { version = "0.4.19", optional = true }
 plotters-svg = {version = "^0.3.*", optional = true}
+stretch = "0.3.2"
```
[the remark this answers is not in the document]
siefkenj: I will lookup how to do this.

examples/layout.rs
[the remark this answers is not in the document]
siefkenj: I didn't intend this example to be *the* example. It's just some sample code. We can put a correct example that properly demonstrates the features when the API has settled.

src/chart/layout/nodes.rs
siefkenj: That might work! I will try it.

src/drawing/area.rs
siefkenj: If I put `#[cfg(feature = "stretch")]` before that line, will it only be included in the enum if that feature is set?

siefkenj:
Thanks for the review! I have a few additional changes related to automatic axis labeling. Do you want those changes pushed to this branch, or to a separate one?

> In general, I think the API workflow is a little bit odd. Currently, if my understand is correct, we need to build a ChartLayout and then pull out the chart_area from the layout then configure the chart context. But could we just build a chart context on top of ChartLayout?

I tried to do this, but I got into a type/lifetime soup and couldn't figure out how to make it work, so I did what was easiest first. Maybe you can help me figure out how to get it working.

In my new work, I have automatic axis labeling work, with space being allocated for the axis labels. I couldn't figure out how to do this with the existing axis format, so I'll need some help figuring that out.

> I am pretty excited about what the current PR have, but I probably have some idea to tweak the workflow later. It looks good to me with the API if we can put it behind a feature switch.

Of course! I fully expect you to tweak things :-)

src/chart/layout/nodes.rs
siefkenj: It doesn't work. `Numeric` is needed to disambiguate between `T: Numeric` and `S: Into<f32>` as used in `(S,S,S,S)`. Since a tuple _could_ implement `Info<f32>`, the type system gets confused. Thats why I needed `Numeric`, which allows me to restrict to an explicit number type.

siefkenj:
I have addressed most of your comments and included the additional code for auto-sizing the label areas (so that tick marks don't fall off the edge of the canvas). Unfortunately, to do that I had to create a new `Axis` type, because I could notfigure out how to extract the necessary information from `Ranged`.

Here's an animation of what it can do:

layout2

Notice how room is automatically allocated for the `.0` on the `3.0`. When the graph starts scrolling, the `.0` doesn't fall off the edge and so the axis runs right up to the edge of the chart instead (it looks like a flaw in the animation, but it actually isn't!)

In order to achieve this, I needed to compute the `extents` of all the tick marks and tick labels. I then union them together and resize the `tick_label_area`s accordingly. If an extent spills to the right/left/top/bottom, I check that area to make sure there is already enough space allocated. If there isn't, I allocate more space.

This process is iterative because, for example, if a very wide left tick label is allocated, then this will affect the width of the x-axis and will potentially affect how many tick marks there are.

Right now, I only implemented everything in f32. You may want to replace the axis stuff with your `Ranged` objects, so I will refrain from making it general, for now.

siefkenj:
For reference, here is what it looks like if a right label also exists. Since adequate space is always allocated, there's no bouncing!

layout2

38:
>  because I could notfigure out how to extract the necessary information from `Ranged`.

Just don't get the point. Could you please what is bouncing in the first animation ? Sorry for the ignorance

siefkenj:
> > because I could notfigure out how to extract the necessary information from `Ranged`.
> 
> Just don't get the point. Could you please what is bouncing in the first animation ? Sorry for the ignorance

On the first GIF, watch the "Sine" label carefully as the number `4.0` scrolls by on the x axis. The `Sine` label is always aligned to the right of the chart. You can see it jump as the `4.0` passes by. That is the layout algorithm making room for the `.0` on the right side.

38:
Ah, I see. But I'm not sure what information you need to avoid this. Probably I need to dig into your code a little bit.

siefkenj:
> Ah, I see. But I'm not sure what information you need to avoid this. Probably I need to dig into your code a little bit.

It's not a bug...it's how it is supposed to work.

The information I need relates to the separate issue of computing the extents of the axis labels. I couldn't figure out how to get the list of labels a `Ranged` was going to print, so I had to make a new struct, `Axis`, where I could figure out which tick labels were going to be printed so I could compute their sizes.

src/chart/layout/mod.rs
```diff
@@ -0,0 +1,995 @@
...
+                .right
+                .as_ref()
+                .map(|range| suggest_tickmark_spacing_for_range(range, h)),
+        };
+        Ok(ret)
+    }
+
+    /// Apply a cartesian grid to the chart area. Major/minor ticks are automatically
+    /// determined on a call to `draw`.
+    pub fn build_cartesian_2d(
```
38, line 619: This function should be generic, because we need to support all types of coords. 

See https://github.com/38/plotters/blob/b82174aa1ac758ca6afb22a555d8732209fb2bb8/src/chart/builder.rs#L179

src/chart/layout/mod.rs
```diff
@@ -0,0 +1,995 @@
...
+
+        // Layouts are cached, so if we didn't change anything, this is a very cheap function call
+        self.nodes.layout(canvas_w as u32, canvas_h as u32)?;
+
+        Ok(())
+    }
+
+    /// Helper function for drawing ticks. This function will call
+    /// `draw_func(pixel_coords, tick, axis_side)` for every tick on every axis.
+    fn draw_ticks_helper<F>(
```
38, line 518: If my understand  was correct, currently the tick is drawn by newly added code instead of the code in ChartContext?

38:
> > Ah, I see. But I'm not sure what information you need to avoid this. Probably I need to dig into your code a little bit.
> 
> It's not a bug...it's how it is supposed to work.
> 
> The information I need relates to the separate issue of computing the extents of the axis labels. I couldn't figure out how to get the list of labels a `Ranged` was going to print, so I had to make a new struct, `Axis`, where I could figure out which tick labels were going to be printed so I could compute their sizes.

See example https://github.com/38/plotters/blob/b82174aa1ac758ca6afb22a555d8732209fb2bb8/src/chart/builder.rs#L179

Use `X:AsRangedCoord` instead of `Range<f32>` for coord spec will allowing you get tick labels. 

```
    x_spec.into().key_points(BoldPoints(n))
```

where `n` is the maximum number of tick marks.

src/chart/layout/mod.rs
siefkenj: That's correct. If I could turn an `AsRangedCoord` into an iterator that iterated through all the tick marks and labels, I could use the existing data structures to do it (this is what I couldn't figure out how to do).

I need to find the extents of each text item/tick mark to compute the correct amount of space.

src/chart/layout/mod.rs
siefkenj: I don't know how to turn `AsRangedCoord` into `Iterator<Tick>`, which is what my code needs to estimate the size of the tick labels.

src/chart/layout/mod.rs
38: Hmm, but I think we probably don't want having two different set of drawing code. If this is the case, I may need to think about unify those code into one. Also, I have the feel that this change seems a little bit big. So I think it's better to split it into parts.

Apologize for this, but I feel we should split the change into parts, since I think we are currently on a way leading to significant refactoring. 

If this is the case, I am going to dig into commit 9938d7b, tweak a little bit and merge it.  I will leave this PR open. After that let's figure out some better way to integrate rest part of the PR properly. 

Please let me know what's your idea.

siefkenj: That sounds good. I can rebase from there after you've tweaked it.

philocalyst:
@siefkenj @38 Any updates in this?

siefkenj:
@philocalyst Unfortunately, this PR has grown stale. Someone else would need to take up the task of rebasing.

philocalyst:
Got it, just was wondering if you had any interest in doing that.

▰▰▰▰▰
Miles Wirth 🙃

Issue #190, opened 2020-09-20
Zero division panic with datetime axis and short span of data


kangalioo:
Hi,

when rendering a plot with a datetime x axis where all data points lie on a single day, plotters crashes because of a divide-by-zero attempt:

```
thread 'serenity client' panicked at 'attempt to divide by zero', /home/kangalioo/.cargo/registry/src/github.com-1ecc6299db9ec823/plotters-0.3.0/src/coord/ranged1d/types/datetime.rs:205:24
stack backtrace:
[snipped]
  13: core::panicking::panic
             at src/libcore/panicking.rs:54
  14: <plotters::coord::ranged1d::types::datetime::RangedDate<D> as plotters::coord::ranged1d::Ranged>::key_points
             at /home/kangalioo/.cargo/registry/src/github.com-1ecc6299db9ec823/plotters-0.3.0/src/coord/ranged1d/types/datetime.rs:205
  15: plotters::coord::ranged2d::cartesian::Cartesian2d<X,Y>::draw_mesh
             at /home/kangalioo/.cargo/registry/src/github.com-1ecc6299db9ec823/plotters-0.3.0/src/coord/ranged2d/cartesian.rs:57
  16: plotters::drawing::area::DrawingArea<DB,plotters::coord::ranged2d::cartesian::Cartesian2d<X,Y>>::draw_mesh::{{closure}}
             at /home/kangalioo/.cargo/registry/src/github.com-1ecc6299db9ec823/plotters-0.3.0/src/drawing/area.rs:199
  17: plotters::drawing::area::DrawingArea<DB,CT>::backend_ops
             at /home/kangalioo/.cargo/registry/src/github.com-1ecc6299db9ec823/plotters-0.3.0/src/drawing/area.rs:277
  18: plotters::drawing::area::DrawingArea<DB,plotters::coord::ranged2d::cartesian::Cartesian2d<X,Y>>::draw_mesh
             at /home/kangalioo/.cargo/registry/src/github.com-1ecc6299db9ec823/plotters-0.3.0/src/drawing/area.rs:198
  19: plotters::chart::context::ChartContext<DB,plotters::coord::ranged2d::cartesian::Cartesian2d<X,Y>>::draw_mesh_lines
             at /home/kangalioo/.cargo/registry/src/github.com-1ecc6299db9ec823/plotters-0.3.0/src/chart/context.rs:170
[...]

PR #279, opened 2021-08-21
Avoid `attempt to divide by zero` when x-axis is datetime

komi1230:
When we use datetime type for x-axis and the data is in the same week, this crate crashes with saying `panicked at 'attempt to divide by zero'`.

This is caused by the calculation of the number of points in a week.
When all data is in a week, `num_weeks()` method returns 0 and results in panic.

close https://github.com/38/plotters/issues/190

### DIFF
--- a/src/coord/ranged1d/types/datetime.rs
+++ b/src/coord/ranged1d/types/datetime.rs
@@ -200,6 +200,12 @@ where
             return ret;
         }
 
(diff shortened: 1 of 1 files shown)


AaronErhardt (approved):

AaronErhardt:
Thanks!

PR #286, opened 2021-09-22
Implemented BitMapWrapper for jupyter notebook

lisanhu:
Implemented a BitMapWrapper for jupyter notebook
function called evcxr_bitmap_figure
inside evcxr.rs

The bitmap is converted to png base64 representation
@38


Cargo.toml
```diff
@@ -95,8 +96,9 @@ fontconfig-dlopen = ["font-kit/source-fontconfig-dlopen"]
 
 # Misc
 datetime = ["chrono"]
-evcxr = ["svg_backend"]
+evcxr = ["svg_backend", "bitmap_backend", "image"]
```
38, line 99: And I think this means once evcxr feature is on then bitmap_backend will be compiled. And I feel this is an additional burden

Cargo.toml
```diff
@@ -95,8 +96,9 @@ fontconfig-dlopen = ["font-kit/source-fontconfig-dlopen"]
...
 deprecated_items = [] # Keep some of the deprecated items for backward compatibility
+evcxr_bitmap_figure = ["bitmap_backend", "evcxr"]
```
38, line 101: And I think want we could do is instead of adding a new feature option, we can use the conditional compilation in the code. I mean if user enables both evcxr and bitmap_backend and then the code for BitMapWrapper will be compiled

src/evcxr.rs
```diff
@@ -39,3 +43,73 @@ pub fn evcxr_figure<
     draw(root).expect("Drawing failure");
     SVGWrapper(buffer, "".to_string())
 }
+
+#[cfg(feature = "evcxr_bitmap_figure")]
```
38, line 47: Here, we can change it to
```
#[cfg(all(feature = "evcxr", feature = "bitmap_backend"))]
```

src/evcxr.rs
```diff
@@ -39,3 +43,73 @@ pub fn evcxr_figure<
...
+pub struct BitMapWrapper(String, String);
+
+#[cfg(feature = "evcxr_bitmap_figure")]
```
38, line 50: Just like I previously mentioned

Cargo.toml
```diff
@@ -16,6 +16,7 @@ exclude = ["doc-template", "plotters-doc-data"]
 num-traits = "0.2.14"
 chrono = { version = "0.4.19", optional = true }
 plotters-svg = {version = "^0.3.*", optional = true}
+base64 = "*"
```
38, line 19: I am a little bit hesitate on this, two things:
first, we should be conservative on introducing new unconditional dependencies. In fact, building time for plotters with minimal feature is one of our major goal. As I discussed with downstream crates (especially criterion), they don't want to introduce additional dependency to build.
Second, you should use a version, instead of star - otherwise, cargo won't allow publish your crate.

Actually, I have a very minimal version base64 implementation available with the SVG backend. So I think it would be nice if you can reuse that. 

I mean create a SVGWrapper and then call blit_bitmap

src/evcxr.rs
```diff
@@ -1,6 +1,10 @@
 use crate::coord::Shift;
 use crate::drawing::{DrawingArea, IntoDrawingArea};
+use base64;
```
38, line 3: As I mentioned previously, I think we should reuse the SVGBackend::blit_bitmap

Though it's a little bit weird that wrap a bitmap in a SVG, but I think this will reduce the number of dependencies we introduced - also reuse the existing code.

38 (changes requested):
Thanks for contribution, I think this is very good overall but I would like to ask for a few changes. Please let me know if you have any thoughts. Thanks!

src/evcxr.rs
lisanhu: Got it! I'll make the change soon!

src/evcxr.rs
lisanhu: That's a good idea! I wasn't aware of the SVG base64 implementation, I'll take a look and make the change! I think it's a good practice to reuse existing code.

lisanhu:
I'll make the change ASAP! Probably sometime next week. Thanks for your comments!

lisanhu:
Hi Hao, sorry for responding late. I just got time to fix the above conflicts. (now it's using blit_bitmap with a svgwrapper).

I noticed one thing here. The blit_bitmap function will convert a bitmap binary to a series of nodes in SVG format. Which is not solving my initial problem of using evcxr_figure and leading me to create evcxr_bitmap_figure.

Using SVG in jupyter notebook is good, it is lightweight in code. However, SVG is sometimes too slow to process in the browsers, and the generated image could not be simply right clicked and saved to local. This is my reason why we need an evcxr_bitmap_figure.

Here I will paste some test code for you
```
:dep plotters = { path = "/Users/lisanhu/mine/tmp/plotters/plotters", version = "*", default_features = false, features = ["evcxr", "bitmap_backend", "all_series"] }
extern crate plotters;
use plotters::prelude::*;
use plotters::evcxr::*;
use std::fs::File;
use std::io::prelude::*;
use plotters::style::text_anchor::{Pos, HPos, VPos};
```
```
let text_style = TextStyle::from(("Courier", 15).into_font())
    .color(&BLACK).pos(Pos::new(HPos::Center, VPos::Center)); // Setting the appropriate text size and font

// Figure is the overall structure containing our rectangles.
let figure = evcxr_bitmap_figure((350, 350), |root| {
    root.fill(&BLACK)?; // Black background fill
    let area = root.margin(40, 10, 40, 10); // Give the main plot some padding on each side.
    area.fill(&WHITE)?; // Fill the main plot area with white color.
    
    for i in 0..100 { // Make a 10x10 plot of 30px by 30px squares.
        let row = i / 10; // Current row index
        let col = i % 10; // Current column index
        let lt = (col * 30, row * 30); // Top left coordinate of this cell
        let rb = (lt.0 + 30, lt.1 + 30); // Bottom right coordinate of this cell
        
    // Create a cyan square from the top left to bottom right coordinates of each cell
        
        area.draw(
            &Rectangle::new([lt, rb], ShapeStyle { color: CYAN.to_rgba(), filled: true, stroke_width: 1 })
        )?;
        
    // Create a Red square of the same size in the same position, with a stroke of 1 and no fill to create padding/
        
        area.draw(
            &Rectangle::new([lt, rb], ShapeStyle { color: RED.to_rgba(), filled: false, stroke_width: 1 })
        )?;
        
    // Draw testing text at the center of each box.
        
        area.draw_text("test", &text_style, (lt.0 + 15, lt.1 + 15))?;
    }
    
    Ok(())
    });
figure

```

After correctly configuring jupyter notebook with evcxr. I noticed the speed of the output figure in my browser is super slow (using base64 representation is super fast). There are also stripes in the black area of the figure found in the rendered image, which is not expected. And we could not right-click on the figure to save it to a bitmap. If we want a SVG image, why do we need to do a bitmap one and convert it? We can directly use the default evcxr_figure.

Currently, I have synced with the upstream code (there seems to be a lot of change in code structures). Remove the base64 dependency, update the compiling conditions, remove the evcxr_bitmap_figure feature, and use blit_bitmap function (all the above conversations are resolved as your suggestions). 

I'm still wishing to get your response about how we should handle the bitmap figure in jupyter notebooks. Do I make mistakes when using blit_bitmap? If I'm using it correctly, shall we keep it like this? Shall we totally remove the evcxr_bitmap_figure from the mainstream codebase? Shall we convert bitmap to base64 representation and embedded it in the browser?

38:
Ah, I see. https://github.com/plotters-rs/plotters/blob/72b2e774f315a134b0c55ea8ce68395f64544b64/plotters-svg/src/svg.rs#L483

This is actually a `plotters-svg` bug, as there's no `image` feature at all - And this should be fixed. 

Since the image encoding feature in `plotters-svg` can not be enabled, so the backend use the default blit_bitmap implementation at there https://github.com/plotters-rs/plotters/blob/72b2e774f315a134b0c55ea8ce68395f64544b64/plotters-backend/src/lib.rs#L294, which is what you are seeing. 

I tried your change it actually works after I feature the plotters-svg feature bug. 

```rust
:dep plotters = { git = "https://github.com/plotters-rs/plotters.git", default_features = false, features = ["evcxr_bitmap", "all_series"] }
use plotters::prelude::*;
use plotters::evcxr::*;
use std::fs::File;
use std::io::prelude::*;
use plotters::style::text_anchor::{Pos, HPos, VPos};

let text_style = TextStyle::from(("Courier", 15).into_font())
    .color(&BLACK).pos(Pos::new(HPos::Center, VPos::Center)); // Setting the appropriate text size and font

// Figure is the overall structure containing our rectangles.
let figure = evcxr_bitmap_figure((350, 350), |root| {
    root.fill(&BLACK)?; // Black background fill
    let area = root.margin(40, 10, 40, 10); // Give the main plot some padding on each side.
    area.fill(&WHITE)?; // Fill the main plot area with white color.
    
    for i in 0..100 { // Make a 10x10 plot of 30px by 30px squares.
        let row = i / 10; // Current row index
        let col = i % 10; // Current column index
        let lt = (col * 30, row * 30); // Top left coordinate of this cell
        let rb = (lt.0 + 30, lt.1 + 30); // Bottom right coordinate of this cell
        
    // Create a cyan square from the top left to bottom right coordinates of each cell
        
        area.draw(
            &Rectangle::new([lt, rb], ShapeStyle { color: CYAN.to_rgba(), filled: true, stroke_width: 1 })
        )?;
        
    // Create a Red square of the same size in the same position, with a stroke of 1 and no fill to create padding/
        
        area.draw(
            &Rectangle::new([lt, rb], ShapeStyle { color: RED.to_rgba(), filled: false, stroke_width: 1 })
        )?;
        
    // Draw testing text at the center of each box.
        
        area.draw_text("test", &text_style, (lt.0 + 15, lt.1 + 15))?;
    }
    
    Ok(())
    });
figure
```

---

I've rebased your code and merged to master. But I don't know why github doesn't show this PR as merged status. 
So I just close this PR, please note this PR is actually merged - please don't mind. 

Again, thanks for the great contribution to plotters!

Cheers,
Hao

lisanhu:
Ah, it's great to know that. Thanks!

PR #290, opened 2021-09-29
Allow filled candle-stick diagrams

### LINKED ISSUE #281: [BUG] CandleStick does not utilize filled Style
**Describe the bug**
When creating a `CandleStick` such as:
```rust
let candle = CandleStick::new(
                    ...,
                    GREEN.filled(),
                    RED.filled(),
                    15,
                );
```
Even though the provided `ShapeStyle` is filled, the graph does not render the rectangle as filled.

**To Reproduce**
Use the current candlestick example and just add `.filled()` to the colors.

**Version Information**
`0.3.1`

**NOTE**
The logic is due to this: https://github.com/38/plotters/blob/72479a086a2f50b4c73d6e613cbf53b7aea604f5/src/element/candlestick.rs#L81
where `false` is always passed as the `fill` parameter for `draw_rect`.

Bromeon:
Implements fill for candle-stick diagrams.
Also uses this functionality in the provided stocks example.

In contrast to #226, this implementation reuses the gain/loss style attribute and does not add an extra field. 
The change is thus *much* simpler.

Fixes #281

### DIFF
--- a/examples/stock.rs
+++ b/examples/stock.rs
@@ -28,7 +28,7 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
 
     chart.draw_series(
         data.iter()
-            .map(|x| CandleStick::new(parse_time(x.0), x.1, x.2, x.3, x.4, &GREEN, &RED, 15)),
+            .map(|x| CandleStick::new(parse_time(x.0), x.1, x.2, x.3, x.4, GREEN.filled(), RED, 15)),
     )?;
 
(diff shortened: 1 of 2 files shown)


Bromeon:
Given https://github.com/38/plotters/issues/333, I don't think this is going anywhere. If there should ever be a fork still interested in this, feel free to contact me.

LeCyberDucky:
Hi!

Given that you only closed this three days ago, I think that https://github.com/38/plotters/issues/341 may be of interest to you.

Issue #207, opened 2020-12-18
[Feature Request] Add pie chart


pachi:
### What is the feature ?
Add pie charts.

### (Optional) Why this feature is useful and how people would use the feature ?

This is usefult to represent the parts in a whole.

### (Optional) Additional Information

Matplotlib has pie chart that could be used as a reference

https://matplotlib.org/3.1.1/gallery/pie_and_polar_charts/pie_features.html

hhamana:
I'm interested in picking this one up. but I'm not sure what  approach can even be possible.
I was originally thinking of making up a triangle for each of those slices, rotate accordingly, and use a circle as mask to round off the edges, but 
- I can't find how to rotate a shape
- looks like using a shape as a mask isn't possible
Another approach would be defining a circular drawing area, and putting the triangles in there to be cropped accordingly, but that's not going to happen either.
Yet again another idea, giving up on circles and triangles, and just looking at building those with polygons, with ideally a curved line for the arc part (Polygon doesn't support curved lines, so that may be at best defined as 1px lines going along the radius?).

I'm just coming up with worse and worse ideas, and even those don't seem to be easily feasible. 
Nevermind allowing something such as exploding a slice, or nested pies.

Can anyone more experienced give some sort of guidance on how that could be implemented?

PR #292, opened 2021-10-23
Pie Element

hhamana:
Implements a high-level element to conveniently draw pie charts (#207 )

This is probably just a first draft implementation, as many parts might benefit from deeper abstraction.
As this works mostly with polar coordinates however, I am not sure which traits would make sense to implement.

Some parts that might be interesting to split off for re-usability, maybe:
- polar coordinates (as ranged2d, along with Cartesian?)
- radians (ranged 1d? std already offers `to_radians()` on f64, not sure what would be the benefit of a newtype).

There are a few flaws, which don't look like plotters offers any better way around:
- This currently draws the arc by calculating a finite number of points, and drawing straight lines in-between. 
Best would be to draw a bezier line that would let me just define the start, middle and end point for each wedge, but I haven't found a way to do so. Due to this, the pie might look aliased at times.
- The backend wants i32 for finite pixels, but the pie works with f64 for radians. Conversion between the two types is lossy. This might create bumps due to coordinates rounding, which doesn't help the first point with aliasing. DrawingBackend should accept floating points coordinates and figure out how to draw smooth half-pixels.

### DIFF
--- a/README.md
+++ b/README.md
@@ -132,6 +132,9 @@ To view the source code for each example, please click on the example image.
 <a href="https://github.com/38/plotters/blob/master/examples/tick_control.rs">
 	<img src="https://plotters-rs.github.io/plotters-doc-data/tick_control.svg" class="galleryItem" width=200px></img>
 </a>
+<a href="https://github.com/38/plotters/blob/master/examples/pie.rs">
+	<img src="https://plotters-rs.github.io/plotters-doc-data/pie_chart.png" class="galleryItem" width=200px></img>
+</a>
 
 
 ## Table of Contents
--- a/examples/pie.rs
+++ b/examples/pie.rs
@@ -15,9 +15,9 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
     let labels = vec!["Pizza", "Pacman"];
 
     let mut pie = Pie::new(&center, &radius, &sizes, &colors, &labels);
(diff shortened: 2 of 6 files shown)


AaronErhardt:

AaronErhardt:
Now that plotters is back, I think you have to revert all those changed links :wink:

38:
Hi, thanks for contribute. This is awesome! Just have a look at the PR, it looks good to me. 

A few comment: 

> * polar coordinates (as ranged2d, along with Cartesian?)

Yes, that is what I was planning to do - but I think this approach is good for now. 

> This currently draws the arc by calculating a finite number of points, and drawing straight lines in-between.
Best would be to draw a bezier line that would let me just define the start, middle and end point for each wedge, but I haven't found a way to do so. Due to this, the pie might look aliased at times.

Yes, this is limited by the backend interface - this requires a change to the backend crate. I think it would be easy to add but I think it's good for now.

> The backend wants i32 for finite pixels, but the pie works with f64 for radians. Conversion between the two types is lossy. This might create bumps due to coordinates rounding, which doesn't help the first point with aliasing. DrawingBackend should accept floating points coordinates and figure out how to draw smooth half-pixels.

Also this has been discussed before - also a backend API limit that we may want to improve in the future. 

Please let me know if you have any questions.

Cheers!

38:
Merged the PR, thanks!

PR #294, opened 2021-10-25
Update documentation

exactly-one-kas:
Update documentation to reflect change from CentricDiscreteRanged to SegmentedCoord

Document a way to create an RGBAColor using RGB and an alpha value


AaronErhardt:

However, I think it makes sense to use the new syntax for doc links with, like ```[`Option`]```: https://doc.rust-lang.org/rustdoc/linking-to-items-by-name.html

PR #297, opened 2021-10-28
Fix some small clippy warnings

remilauzier:


AaronErhardt:

I'd like to merge this without the addition of clippy.toml because this exists now: https://doc.rust-lang.org/cargo/reference/manifest.html#the-rust-version-field

PR #307, opened 2021-11-18
Fix IntoMonthly doc comment

wcassels:


AaronErhardt:

PR #313, opened 2021-12-05
Add derive(Clone) to Cartesian3d and DualCoordChartState

dmeklund:
Currently, attempting to clone a ChartState fails if either a) the ChartState is from a ChartContext that was built with a 3D Cartesian coordinate system, or b) the ChartState is from a ChartContext that was converted into a dual axis chart context. In both cases, adding the `#[derive(Clone)]` annotation to the corresponding class (`Cartesian3d` and `DualCoordChartState` respectively) resolves the issue. Since dual axis chart contexts and 3D cartesian coordinates are already being used in existing tests in `src/chart/context.rs`, I went ahead and added a line to check that chart contexts could be successfully cloned in both cases.

(This bug was discovered while attempting to adapt the `minifb-demo` example to use a 3D Cartesian coordinate system.)

### DIFF
--- a/src/chart/context.rs
+++ b/src/chart/context.rs
@@ -132,6 +132,10 @@ mod test {
             .draw()
             .expect("Draw Secondary axes");
 
+        // test that chart states work correctly with dual coord charts
+        let cs = chart.into_chart_state();
+        let mut chart = cs.clone().restore(&drawing_area);
+
         chart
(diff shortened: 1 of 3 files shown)


Titaniumtown:
Yes! Please merge. I've had this issue myself.

dmeklund:
Thanks for the heads up, will do

38:
Merged, thanks!

PR #315, opened 2021-12-16
Bitmap examples: call present

ijackson:
See https://github.com/38/plotters/issues/189#issuecomment-996174342


AaronErhardt:

PR #337, opened 2022-03-29
update font-kit to 0.11.0 and add fontconfig-dlopen feature

Be-ing:


Be-ing:
Hi @38 are you still maintaining this crate? I don't see any new commits in the last 6 months.

PR #339, opened 2022-04-09
implement DiscreteRanged for RangedDateTime

rksm:
This implements `DiscreteRanged` for `RangedDateTime` so that a date time range can be used for reverse coordinate mapping.

In particular, I stumbled across this when following the [wasm-demo](https://github.com/38/plotters/tree/master/examples/wasm-demo) whose plot functions return a reverse mapper (very useful!). `ChartContext::into_coord_trans` however does not work when using a DateTime range. This addition fixes that.

### DIFF
--- a/src/coord/ranged1d/types/datetime.rs
+++ b/src/coord/ranged1d/types/datetime.rs
@@ -655,6 +655,30 @@ where
     }
 }
 
+impl<DT> DiscreteRanged for RangedDateTime<DT>
+where
+    DT: Datelike + Timelike + TimeValue + Clone + PartialOrd,
+    DT: Add<Duration, Output = DT>,
+    DT: Sub<DT, Output = Duration>,
+    RangedDate<DT::DateType>: Ranged<ValueType = DT::DateType>,
+{
+    fn size(&self) -> usize {
+        ((self.1.clone() - self.0.clone()).num_milliseconds().max(-1) + 1) as usize
+    }
+
+    fn index_of(&self, value: &DT) -> Option<usize> {
+        let ret = (value.clone() - self.0.clone()).num_milliseconds();
+        if ret < 0 {
+            return None;
+        }
+        Some(ret as usize)
+    }
+
+    fn from_index(&self, index: usize) -> Option<DT> {
+        Some(self.0.clone() + Duration::milliseconds(index as i64))
+    }
+}
+
 /// The coordinate that for duration of time
 #[derive(Clone)]
 pub struct RangedDuration(Duration, Duration);


rksm:
Is there anything wrong with that change? Happy to address any concerns.

38:
Hi there, sorry for the delayed reply.

 I looked into the PR, it seems that the PR defines treat 1ms as a unit - this is very limited.

Also, I think what the PR is implementing is already possible with the plotters v0.3.1.

```rust

use plotters::prelude::*;
use chrono::{Utc, TimeZone, Duration};
fn main() {
    let back = SVGBackend::new("x.svg", (1024, 768));
    let mut root = back.into_drawing_area();

    let dt_range = Utc.ymd(2022, 1, 1).and_hms(0,0,0)..Utc.ymd(2022,1,1).and_hms(0,0,1);

    let mut c = ChartBuilder::on(&root)
        .set_left_and_bottom_label_area_size(50)
        .build_cartesian_2d(dt_range.step(Duration::milliseconds(1)), 0u32..10)
        .unwrap();

    c.configure_mesh().draw().unwrap();

    c.into_coord_trans();

}
```

Please let me know if this is not exactly what the PR is for.

Thanks!

rksm:
Hello, thanks for looking at this.

Here is the example that fails with plotters v0.3.1: https://github.com/rksm/plotters-canvas-test/blob/master/src/main.rs#L23

Running `cargo build --target wasm32-unknown-unknown` results in:

```
error[E0599]: the method `into_coord_trans` exists for struct `ChartContext<'_, CanvasBackend, Cartesian2d<RangedDateTime<chrono::DateTime<chrono::Utc>>, RangedCoordi32>>`, but its trait bounds were not satisfied
  --> src/main.rs:58:14
   |
58 |     Ok(chart.into_coord_trans())
   |              ^^^^^^^^^^^^^^^^ method cannot be called on `ChartContext<'_, CanvasBackend, Cartesian2d<RangedDateTime<chrono::DateTime<chrono::Utc>>, RangedCoordi32>>` due to unsatisfied trait bounds
   |
  ::: /Users/robert.krahn/.cargo/registry/src/github.com-1ecc6299db9ec823/plotters-0.3.1/src/coord/ranged2d/cartesian.rs:20:1
   |
20 | pub struct Cartesian2d<X: Ranged, Y: Ranged> {
   | -------------------------------------------- doesn't satisfy `_: ReverseCoordTranslate`
   |
   = note: the following trait bounds were not satisfied:
           `Cartesian2d<RangedDateTime<chrono::DateTime<chrono::Utc>>, RangedCoordi32>: ReverseCoordTranslate`
```

The MR fixes this.

> I looked into the PR, it seems that the PR defines treat 1ms as a unit - this is very limited.

Not quite sure what you mean. The MR implements the `RangedDateTime` on a millisecond resolution which seems reasonable to me. A nanosecond resolution might be an option.

Best,
Robert

38:
Hi Robert, sorry for the confusion. Two things,

First, the DiscreteRanged trait is particularly for the types that has well-defined discrete value on it, but DateTime doesn't have one conceptually. Also the PR only allows 1ms resolution, which may not be very useful in many use cases.

Second, there is more universal way to convert a continuous axis to discrete ones. As my example shown before, which means what you PR is implementing seems to equivalent to just call ```x_range.step(Duration::milliseconds(1))```

See  https://docs.rs/plotters/0.3.1/plotters/coord/combinators/trait.IntoLinspace.html#method.step for details about this API. And this is already living in v0.3.1. Hope this is helpful

Thanks!

38:
Just FYI, your example can be fixed in following way - You can also adjust the resolution by passing different duration to `step` method. 

```rust
use anyhow::Result;
use chrono::{prelude::*, Duration};
use plotters::prelude::*;
use plotters_canvas::CanvasBackend;
use wasm_bindgen::JsCast;
use web_sys::HtmlCanvasElement;

fn main() {
    console_error_panic_hook::set_once();

    let canvas = web_sys::window()
        .unwrap()
        .document()
        .unwrap()
        .get_element_by_id("canvas")
        .unwrap()
        .dyn_into::<HtmlCanvasElement>()
        .unwrap();

    let _tfm_fn = render_plot(canvas).unwrap();
}

fn render_plot(
    canvas: HtmlCanvasElement,
) -> Result<impl Fn((i32, i32)) -> Option<(DateTime<Utc>, i32)>> {
    let backend = match CanvasBackend::with_canvas_object(canvas) {
        None => return Err(anyhow::anyhow!("cannot init plotters canvas backend")),
        Some(backend) => backend,
    };

    let draw_area = backend.into_drawing_area();
    draw_area.fill(&WHITE)?;

    let x_range = Utc::now()..(Utc::now() + Duration::minutes(10));
    let y_range = 0i32..10i32;
    let data = (0..=10).map(|n| (Utc::now() + Duration::minutes(n as i64), n as i32));

    let mut chart = ChartBuilder::on(&draw_area)
        .margin(25u32)
        .caption("test plot", FontDesc::from(("sans-serif", 12.0)))
        .x_label_area_size(30u32)
        .y_label_area_size(30u32)
        .build_cartesian_2d(x_range.step(Duration::milliseconds(1)), y_range)?;

    chart.draw_series(LineSeries::new(data, RED))?;

    chart
        .configure_mesh()
        .x_labels(4)
        .y_labels(4)
        .x_label_formatter(&|x| x.format("%H:%M:%S %.3f").to_string())
        .draw()?;

    draw_area.present()?;

    Ok(chart.into_coord_trans())
}

```

rksm:
Thank you for providing the detailed explanation, much appreciated. I see your point now.

Issue #169, opened 2020-07-21
More shorthand functions and ergonomics


kangalioo:
### What is the feature ?
I'm trying out this library for the first time and there's some places where it would be delightful to have shorthand functions:
1. Shorthand functions for Pos: `Pos::top_right`, `Pos::bottom_right`, `Pos::bottom_center` etc..
2. Short way to set colors for components. Currently, lots of boilerplate is required for e.g. white-on-black plotos
    - a cool start would be to have IntoTextStyle implementation for `(FontFamily, T, RGBAColor)`
3. A direct way to create an RGBAColor
4. A more sensible default font size than one pixel [see here](https://docs.rs/plotters/0.2.15/src/plotters/style/font/font_desc.rs.html#128-132)
5. Common trait derives for RGBColor, i.e. Copy, Clone, Eq, PartialEq, Default, Hash
6. `From<RGBColor>` impl for RGBAColor
7. Expose BitMapBackendError to allow handling those errors

### Why this feature is useful and how people would use the feature ?
More shorthand functions and ergonomics are always good, I think

38:
Thanks for the suggestions and sorry for the delay. I haven't got a chance to look at the issue since a few month ago.

Most of suggestion seems really helpful, but I would like to understand more about few things: 


>     1. Shorthand functions for Pos: `Pos::top_right`, `Pos::bottom_right`, `Pos::bottom_center` etc..

What is this particularly for ? Since we have positioning enums used in both ChartBuilder and legends, which one are you referring or both?

Thanks!

38:
Just put the items here for tracking purposes. 


- [ ] Short way to set colors for components: e.g. IntoTextStyle implementation for `(FontFamily, T, RGBAColor)`

- [ ]  A direct way to create an RGBAColor
- [ ]  Default font size? (This might be a chance to revisit #73 ) 
- [ ] RGBColor should impl more traits 
- [ ] From<RGBColor> for RGBAColor
- [ ] BitmapBackendError visibility

kangalioo:
> > ```
> > 1. Shorthand functions for Pos: `Pos::top_right`, `Pos::bottom_right`, `Pos::bottom_center` etc..
> > ```
> 
> What is this particularly for ? Since we have positioning enums used in both ChartBuilder and legends, which one are you referring or both?

Looking back at it, I think I had in mind a shorthand for `Pos::new(HPos::XXX, VPos::XXX)`. So,

[...]

PR #359, opened 2022-06-06
Derive common traits for HSLColor and PaletteColor

### LINKED ISSUE #340: [Feature Request] Derive Copy for HSLColor
### What is the feature ?

Please, derive `Copy` (and, by extension, `Clone`) for `HSLColor`.

### Why this feature is useful and how people would use the feature ?

Currently, manipulating `HSLColor` is a burden, specially if it happens to be the type of a field of some struct. My entire struct looses the ability to be copied (or even cloned!). And it's completely unnecessary, given that `HSLColor` is a simple `struct` with just 3 `f64`.

Thanks in advance!

facorread:
Close #340
Addresses #169

Deriving common traits like `Clone`, `Copy`, `PartialEq`, `Debug`, and `Default` for structs `HSLColor` and `PaletteColor`. Similar work was previously done for `RGBColor` and `RGBAColor` in commit 8866f9e.

### DIFF
--- a/src/style/color.rs
+++ b/src/style/color.rs
@@ -77,6 +77,7 @@ impl Color for RGBAColor {
 }
 
 /// A color in the given palette
+#[derive(Copy, Clone, Eq, PartialEq, Hash, Debug, Default)]
 pub struct PaletteColor<P: Palette>(usize, PhantomData<P>);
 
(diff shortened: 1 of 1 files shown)


AaronErhardt (approved):

AaronErhardt:
Thanks!

PR #360, opened 2022-06-07
Minor gridline density in 2d and 3d

### LINKED ISSUE #303: [Feature Request] Minor gridline density
### What is the feature ?
Allow the user to specify the amount of light/minor grid lines

### Why this feature is useful and how people would use the feature ?
When creating a graph that is small or has many major grid lines close together, it may be ideal for the amount of minor grid lines to decrease to less than the currently hard coded, and quite aggressive 10.

### Additional Information
The implementation looks to have the amount of minor grid lines hard coded here: 
https://github.com/38/plotters/blob/master/src/chart/mesh.rs#L444-L445

#### Example
Ignoring the otherwise messed up formatting, the 10 minor grid lines basically join together to create a solid background, which is not what is wanted.

image

facorread:
The 2d grid can be seen on the tests at `sample.png`; the 3d grid can be seen on the tests at `3d-plot.svg` and `3d-plot2.gif`.

sample

3d-plot

3d-plot2

Close #303

### DIFF
--- a/src/chart/axes3d.rs
+++ b/src/chart/axes3d.rs
@@ -45,28 +45,28 @@ where
         self
     }
 
-    /// Set the maximum number of divisions for the fine grind grid
+    /// Set the maximum number of divisions for the minor grid
     /// - `value`: Maximum desired divisions between two consecutive X labels
     pub fn x_max_light_lines(&mut self, value: usize) -> &mut Self {
         self.light_lines_limit[0] = value;
         self
     }
 
-    /// Set the maximum number of divisions for the fine grind grid
+    /// Set the maximum number of divisions for the minor grid
     /// - `value`: Maximum desired divisions between two consecutive Y labels
     pub fn y_max_light_lines(&mut self, value: usize) -> &mut Self {
         self.light_lines_limit[1] = value;
         self
     }
 
-    /// Set the maximum number of divisions for the fine grind grid
+    /// Set the maximum number of divisions for the minor grid
     /// - `value`: Maximum desired divisions between two consecutive Z labels
     pub fn z_max_light_lines(&mut self, value: usize) -> &mut Self {
         self.light_lines_limit[2] = value;
         self
     }
 
-    /// Set the maximum number of divisions for the fine grind grid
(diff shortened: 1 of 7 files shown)


facorread:
Close #303

src/chart/axes3d.rs
```diff
@@ -130,9 +162,9 @@ where
             BoldPoints(self.n_labels[2]),
         );
         let kps_light = chart.get_key_points(
-            LightPoints::new(self.n_labels[0], self.n_labels[0] * 10),
-            LightPoints::new(self.n_labels[1], self.n_labels[1] * 10),
-            LightPoints::new(self.n_labels[2], self.n_labels[2] * 10),
+            LightPoints::new(self.n_labels[0] * self.light_lines_limit[0], self.n_labels[0] * self.light_lines_limit[0]),
+            LightPoints::new(self.n_labels[1] * self.light_lines_limit[1], self.n_labels[1] * self.light_lines_limit[1]),
+            LightPoints::new(self.n_labels[2] * self.light_lines_limit[2], self.n_labels[2] * self.light_lines_limit[2])
```
AaronErhardt, line 167: This looks odd to me, shouldn't it be

```rust
LightPoints::new(self.n_labels[0], self.n_labels[0] * self.light_lines_limit[0]),
```

Otherwise we would be multiplying both arguments which wasn't the case before and also doesn't happen in the 2d version.

src/chart/mesh.rs
```diff
@@ -308,6 +312,29 @@ where
         self.axis_style = Some(style.into());
         self
     }
+
+    /// Set the maximum number of divisions for the fine grind grid
```
AaronErhardt, line 316: I'm not a native speaker, but I've never heard of a "grind grid". Maybe this can be described better.

AaronErhardt (changes requested):
Looks good to me overall, but I'm not an expert regarding plotters internals. Ideally someone else could have a look whether the naming makes sense. Maybe `max_grid_lines` or `max_light_grid_lines` is more intuitive.

And the rustfmt job fails which needs to be fixed.

PR #362, opened 2022-06-09
Bump `image` version `0.23.13` -> `0.24.2`

### LINKED ISSUE #361: BitMapElement error with latest `image` crate
I'm using the latest `image` crate ([`0.24.2`](https://docs.rs/image/latest/image/imageops/colorops/index.html)) and I'm trying to do something similar to the [`blit-bitmap`](https://github.com/plotters-rs/plotters/blob/master/examples/blit-bitmap.rs#L34) example. 

In particular, my compilation fails at:

```rust
let elem: BitMapElement<_> = ((0.05, 0.95), image).into();
``` 

The error is the following:

```
the trait bound `plotters::element::BitMapElement<'_, _>: From<(({float}, {float}), DynamicImage)>` is not satisfied
```

If I set the version to image `0.23.14` it works ok, but this would mean losing quite a lot of useful features from the latest release.

Is there a plan to support image `^0.24.*` anytime soon?

arturomf94:
Closes #361

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -26,7 +26,7 @@ jobs:
           submodules: recursive
     - uses: actions-rs/toolchain@v1
       with:
-          toolchain: 1.46.0
+          toolchain: 1.56.0
           override: true
     - uses: actions-rs/cargo@v1
       with:
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -32,7 +32,7 @@ pathfinder_geometry = { version = "0.5.1", optional = true }
 font-kit = { version = "0.11.0", optional = true }
 
 [target.'cfg(not(target_arch = "wasm32"))'.dependencies.image]
-version = "0.23.14"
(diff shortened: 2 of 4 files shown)


AaronErhardt (changes requested):
Looks good to me, but we'd have to increase the MSRV to 1.56.0 it seems to support the newer image crate. From my side there's nothing speaking against that, but ideally we should also adjust the Rust edition of plotters to "2021", the MSRV in the README and the MSRV CI job. Could you add this to your changes?

arturomf94:
> Looks good to me, but we'd have to increase the MSRV to 1.56.0 it seems to support the newer image crate. From my side there's nothing speaking against that, but ideally we should also adjust the Rust edition of plotters to "2021", the MSRV in the README and the MSRV CI job. Could you add this to your changes?

Sure! Just did :)

AaronErhardt (approved):
Perfect, thanks!

PR #364, opened 2022-06-12
Shorten some lifetimes to prevent borrowing errors. Close #363

### LINKED ISSUE #363: [BUG] minifb example broken
I'm trying to run the minifb example (https://github.com/plotters-rs/plotters/blob/master/examples/minifb-demo/src/main.rs) , but I get this error:
```
error[E0499]: cannot borrow `buf` as mutable more than once at a time
   --> src/main.rs:121:17
    |
73  |         BitMapBackend::<BGRXPixel>::with_buffer_and_format(buf.borrow_mut(), (W as u32, H as u32))?
    |                                                            ---------------- first mutable borrow occurs here
...
121 |                 buf.borrow_mut(),
    |                 ^^^^^^^^^^^^^^^^ second mutable borrow occurs here
...
187 | }
    | - first borrow might be used here, when `root` is dropped and runs the destructor for type `DrawingArea<BitMapBackend<'_, BGRXPixel>, Shift>`

error[E0502]: cannot borrow `buf` as immutable because it is also borrowed as mutable
   --> src/main.rs:175:39
    |
73  |         BitMapBackend::<BGRXPixel>::with_buffer_and_format(buf.borrow_mut(), (W as u32, H as u32))?
    |                                                            ---------------- mutable borrow occurs here
...
175 |             window.update_with_buffer(buf.borrow())?;
    |                                       ^^^^^^^^^^^^ immutable borrow occurs here
...
187 | }
    | - mutable borrow might be used here, when `root` is dropped and runs the destructor for type `DrawingArea<BitMapBackend<'_, BGRXPixel>, Shift>`

Some errors have detailed explanations: E0499, E0502.
For more information about an error, try `rustc --explain E0499`.
```

Now I'm new to rust, but I don't get how this is supposed to work, since it the first mutable of borrow of `buf` wont go out of scope any time in main, so no other function will be able to borrow it, or what am I missing?

facorread:
Shorten some lifetimes to prevent borrowing errors in the `minifb` example.

Close #363

### DIFF
--- a/examples/minifb-demo/Cargo.toml
+++ b/examples/minifb-demo/Cargo.toml
@@ -7,6 +7,6 @@ edition = "2018"
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [dependencies]
-minifb = "0.13.0"
+minifb = "0.23.0"
 plotters = { path = "../.." , default_features = false, features = ["ttf"]}
 plotters-bitmap = {version = "^0.3.*", default_features = false}
--- a/examples/minifb-demo/src/main.rs
+++ b/examples/minifb-demo/src/main.rs
@@ -69,24 +69,27 @@ fn main() -> Result<(), Box<dyn Error>> {
         H,
         WindowOptions::default(),
     )?;
-    let root =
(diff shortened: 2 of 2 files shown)


AaronErhardt:
Thanks!

PR #367, opened 2022-06-15
Remove deprecation warnings in benchmark

shinmili:
`Criterion::bench` and `ParametrizedBenchmark` is now marked as
deprecated. This patch replaces them with a new `BenchmarkGroup` API.

### DIFF
--- a/benches/benches/parallel.rs
+++ b/benches/benches/parallel.rs
@@ -1,4 +1,4 @@
-use criterion::{criterion_group, Criterion, ParameterizedBenchmark};
+use criterion::{criterion_group, BenchmarkId, Criterion};
 
 use plotters::coord::Shift;
 use plotters::prelude::*;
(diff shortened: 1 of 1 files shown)


facorread:
From `cargo bench`:

WARNING: HTML report generation will become a non-default optional feature in Criterion.rs 0.4.0.
This feature is being moved to cargo-criterion (https://github.com/bheisler/cargo-criterion) and will be optional in a future version of Criterion.rs. To silence this warning, either switch to cargo-criterion or enable the 'html_reports' feature in your Cargo.toml.

AaronErhardt (approved):
Looks good. Thanks!

@facorread I think we'll take of criterion 0.4 and "html_reports" when it's released. I don't think there's anything we can do for now.

shinmili:
Or you can use [`cargo criterion` CLI](https://crates.io/crates/cargo-criterion). Enabling `criterion`'s optional features may lead to longer compilation time for benchmark binaries, and one of the CLI's goals is to reduce it.

PR #373, opened 2022-06-20
Proposal to use git-master dependencies for collaboration, testing, and CI

facorread:
This proposal aims to improve collaboration, testing, and CI by altering the version of the `plotters-backend`, `plotters-bitmap`, and `plotters-svg` dependencies in `Cargo.toml`. Instead of using the stable versions of those crates, the git-master versions are used.

Using the `git-master` versions of the `plotters-` crates is vital for pre-release quality assurance and frustration-free collaboration in Plotters. It also reduces surprises at release time.

Note: [crates.io](https://crates.io/) does not allow packages to be published with git dependencies. See [The Cargo Book](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html) for more information. `Cargo.toml` should be updated to the stable versions right before a new release.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,15 +15,21 @@ exclude = ["doc-template", "plotters-doc-data"]
 [dependencies]
 num-traits = "0.2.14"
 chrono = { version = "0.4.19", optional = true }
-plotters-svg = {version = "^0.3.*", optional = true}
 
 [dependencies.plotters-backend]
-version = "^0.3"
+git = "https://github.com/plotters-rs/plotters-backend"
+# version = "0.3.1"
 
 [dependencies.plotters-bitmap]
-version = "^0.3"
-optional = true
 default_features = false
+git = "https://github.com/plotters-rs/plotters-bitmap"
+optional = true
+# version = "0.3.1"
+
+[dependencies.plotters-svg]
+git = "https://github.com/plotters-rs/plotters-svg"
+optional = true
+# version = "0.3.1"
 
 [target.'cfg(not(target_arch = "wasm32"))'.dependencies]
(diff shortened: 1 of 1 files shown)


Cargo.toml
```diff
@@ -15,15 +15,24 @@ exclude = ["doc-template", "plotters-doc-data"]
 [dependencies]
 num-traits = "0.2.14"
 chrono = { version = "0.4.19", optional = true }
-plotters-svg = {version = "^0.3.*", optional = true}
 
 [dependencies.plotters-backend]
-version = "^0.3"
+branch = "master"
```
AaronErhardt, line 20: Do we need `branch = "master"`? That should work by default, I think.

AaronErhardt (commented):
I agree with the proposal, it should have been like this in the first place. Ideally, we can fix all CI failures before merging.

facorread:
Great! I addressed all the issues locally. However, the Github CI will continue to fail until the PR has been merged for all `plotters-` crates.

38:
> Great! I addressed all the issues locally, so `cargo build --all-targets` is working well for each and every `plotters-` crate at home. However, Github Actions will continue to fail until the PR has been merged for all `plotters-` crates. It's a catch-22 situation, sorry!

Yes, we need to update all the repo at the same time - as the test isn't currently using same `plotters-backend` crate

AaronErhardt:
> I am thinking about do we want to have a repo fore plotters core and some essential backend and dependencies. I mean plotters, plotters-backend, plotters-bitmap and plotters-svg.

That'd be a good step in my opinion. It's rather tedious to have issues and code of closely related crates scattered at different places and other projects like tokio or Relm4 also favor the "monorepo" approach.

PR #374, opened 2022-06-22
Mono repo

38:
Spent some time make a monorepo, just want to see how you guys think about this


AaronErhardt (approved):
Looks good to me. Potentially, we don't need explicit dependabot configs because AFAIK it works with cargo workspaces too.

facorread:
This is a great idea! I would like to suggest, now that you are trying to work out the CI problems, to add the --all-targets argument to the msrv job, just for thoroughness.

facorread (approved):

PR #378, opened 2022-06-25
Fix rect bottom right edges

### LINKED ISSUE #368: [BUG] `DrawingArea::fill` doesn't fill the right and bottom edge in some backends
**Describe the bug**
`DrawingArea::fill` doesn't fill the right and bottom edge in some backends and leaves transparent edges in a generated image. As far as I have checked, [plotters-svg](https://github.com/plotters-rs/plotters-svg) and [plotters-canvas](https://github.com/plotters-rs/plotters-canvas) have this issue. `BitMapBackend` is fine about this.

For CanvasBackend, it's quick to see #165's attached image. It continues in the latest code too.

For SVGBackend, The repro code below prints this result. `width` and `height` in the `rect` element should be `"100"` here.

```svg
<svg width="100" height="100" viewBox="0 0 100 100" xmlns="http://www.w3.org/2000/svg">
<rect x="0" y="0" width="99" height="99" opacity="1" fill="#0000FF" stroke="none"/>
</svg>
```

Other backends with sub-pixel coordinate might have the same issue. Can be related to #128.

**To Reproduce**
```rust
use plotters::prelude::*;

    let mut content = String::new();
    {
        let area = SVGBackend::with_string(&mut content, (100, 100)).into_drawing_area();
        area.fill(&BLUE)?;
    }
    println!("{content}");
    Ok(())
}
```

**Version Information**
`plotters`: current master (a72bfbced3523b17adc974ec7459f6c1c6f2d986)
`plotters-svg`: current master (https://github.com/plotters-rs/plotters-svg/commit/47289e89a1078f330fcb6a3886b0ae34dc3f69e2)
`plotters-canvas`: current master (https://github.com/plotters-rs/plotters-canvas/commit/111fce55a9ccc2999864358f991a28df6261f921)

shinmili:
Close #368.

### DIFF
--- a/.github/workflows/plotters-core.yml
+++ b/.github/workflows/plotters-core.yml
@@ -63,7 +63,7 @@ jobs:
     - uses: actions-rs/cargo@v1
       with:
           command: test
-          args: --verbose --no-default-features --lib
+          args: --verbose --no-default-features --features=svg_backend --lib
   test_all_features:
     runs-on: ubuntu-latest
     steps:
--- a/plotters-bitmap/src/bitmap.rs
+++ b/plotters-bitmap/src/bitmap.rs
@@ -252,12 +252,12 @@ impl<'a, P: PixelFormat> DrawingBackend for BitMapBackend<'a, P> {
         if (from.0 == to.0 || from.1 == to.1) && style.stroke_width() == 1 {
             if alpha >= 1.0 {
                 if from.1 == to.1 {
-                    P::fill_rect_fast(self, from, to, r, g, b);
+                    P::fill_rect_fast(self, from, (to.0 + 1, to.1 + 1), r, g, b);
                 } else {
                     P::fill_vertical_line_fast(self, from.0, (from.1, to.1), r, g, b);
                 }
             } else {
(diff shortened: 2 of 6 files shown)


38:
Thanks for the PR, I think this make a lot of sense to me!

But again, IMO it's not a really a solution but a mitigation - And I am looking forward we can reach the point that we define a sound and uniformed backend abstraction at some point. 

The change itself looks good to me! 

I am wondering if @facorread  and @AaronErhardt have any comment on the PR.

facorread:
I agree! I know it can be frustrating that a project chooses a long-term abstraction over a short-term solution, but the discussion here highlights the need for it. Let us try to build it!

AaronErhardt (approved):
Looks good to me

facorread (changes requested):
This is nitpick of course, can you please address the warning below? It looks like removing one line should do it. Thanks!

shinmili:
@facorread Thanks for pointing out. The warning indicates the test below is actually not run on GitHub Actions, so I edited the test command line instead of removing the line with warning.

facorread (approved):

PR #404, opened 2022-08-08
Bump chrono to 0.4.20

pinkforest:
Chrono now - 0.4.20 - has localtime_r RIR that has addressed the past security issue around it :partying_face: 

We've updated here:
https://rustsec.org/advisories/RUSTSEC-2020-0159.html


AaronErhardt (approved):
Thanks!

Issue #413, opened 2022-09-08
[BUG] `doc!` macro not exported (needed for `define_color!`.


hpux735:
**Describe the bug**
I was working on some custom palettes, and I looked how they were made in the upstream code.  It seems like `define_color!` is the way to do it.  I tried using it myself, but the compiler complained that it couldn't access the `doc!` macro.

**To Reproduce**
Try to make a color yourself:

```
use plotters::define_color;

define_color!(MOODY_BLUE, 0, 63, 92, "*Moody Blue*");

fn main() { }
```

the compiler will complain:
```
error[E0432]: unresolved import `plotters::doc`
 --> src/bin/plotters.rs:7:30
  |
7 | use plotters::{define_color, doc};
  |                              ^^^ no `doc` in the root

error: cannot determine resolution for the macro `doc`
  --> src/bin/plotters.rs:10:1
   |
10 | define_color!(MOODY_BLUE, 255, 255, 255, "*Moody Blue*");
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   |
   = note: import resolution is stuck, try simplifying macro imports
   = note: this error originates in the macro `define_color` (in Nightly builds, run with -Z macro-backtrace for more info)
```

**Version Information**
`Compiling plotters v0.3.3`

The fix is as easy as adding `#[macro_export]` on [line 6](https://github.com/plotters-rs/plotters/blob/master/plotters/src/style/colors/mod.rs#L6)

AaronErhardt:
Closed via #419

PR #419, opened 2022-09-19
Export the `doc` macro, to use of `define_color`

hpux735:
This change exports the `doc` macro, which is required to be exported for the already exported `define_color` macro to work.  This has been documented in bug #413


AaronErhardt (approved):
Thanks! Sorry for the long delay.

PR #420, opened 2022-09-22
Fix spelling and grammar in README.md

novedevo:
Nothing that changes meaning, just minor edits


AaronErhardt:
Thanks a lot for your PR! 

Unfortunately, the actual README lives [here](https://github.com/plotters-rs/plotters/blob/master/doc-template/readme.template.md) as a template. This means that changes to the README will be overwritten eventually. On the other side, changing the the template will improve the docs and the README at the same time.

Honestly, I don't think this is a great solution (and quite hidden as well) but for now that's how it works...

So would you mind porting your changes to the template? If not that's fine too, I will take care of it then :)

novedevo:
All my changes have been ported to the templates. Sorry for missing them the first time through!

AaronErhardt (approved):
Awesome, thanks!

Issue #423, opened 2022-10-09
[Feature Request] Document using the `ab_glyph` feature


Monadic-Cat:
### What is the feature ?
Recently, #311 was merged, and added a new text rendering backend feature.
It would be nice if there were documented instructions on how to use it, especially when the docs.rs page for `plotters` does not show the extra function exported for the purpose (because `cfg` work like that, lol).

### Why this feature is useful and how people would use the feature ?
While it is now possible to use `ab_glyph` for text rendering in `plotters`, I can't find anywhere that is explained outside of browsing the project source code. It would be really nice if the feature were called out, especially for cross-compiling use cases where it's much easier to use `ab_glyph`, as opposed to manually putting together a cross sysroot to make font-config cross compile correctly.

### Additional Information
I'd be happy to write this documentation if someone could tell me what places would be most appropriate for it.

PR #424, opened 2022-10-11
update font backend documentation

Monadic-Cat:
This addresses #423, by updating the font features in the README with `ab_glyph`.
(Also, `rusttype` appears to no longer be in use.)

I am unclear on how to propagate the change on the README template to the actual README.

EDIT: I also added a section to `register_font`'s docs.


AaronErhardt (approved):
Thanks!

PR #427, opened 2022-10-21
Fix broken link of example in README.md

eval-exec:
The console.rs example link in README.md is broken, because plotter now use cargo workspace.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,4 @@
 **/target
 .vscode/*
 Cargo.lock
+.idea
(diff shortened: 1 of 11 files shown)


facorread:
Thanks for your contribution!

facorread (approved):

.gitignore
```diff
@@ -3,3 +3,4 @@
 **/target
 .vscode/*
 Cargo.lock
+.idea
```
AaronErhardt, line 6: Is there any reason for adding this?

AaronErhardt (changes requested):
Thanks for this PR!

Since we're already fixing links, it makes sense to also point to the new repo (https://github.com/plotters-rs/plotters) instead of (https://github.com/38/plotters). Both links will work because the latter will be redirected, but this might break eventually.

.gitignore
eval-exec: `.idea` is the Jetbrains IDE generated folder, so add it to `.gitignore`

AaronErhardt: Alright, makes sense :+1:

AaronErhardt (approved):
Thanks!

PR #430, opened 2022-10-26
Fixed cpustat example link

ansible42:
Updated dead link to correctly point to the [cpustat.rs ](https://github.com/plotters-rs/plotters-piston/blob/master/examples/cpustat.rs)


AaronErhardt (approved):
Thanks!

PR #441, opened 2022-12-23
Make error type 'static.

raldone01:
This pull request makes the drawing backend error type 'static which makes returning errors more pleasant.
Currently the `draw_chart` function was not callable if this was not the case any ways.

```rust
fn some_func<DB>(
    area: &DrawingArea<DB, plotters::coord::Shift>,
  ) -> Result<(), Box<dyn std::error::Error>>
  where
    DB: DrawingBackend,
    // DB::ErrorType: 'static, not necessary any more, not very intuitive
  {
    let mut chart = ChartBuilder::on(area)
      .margin(10)
      .x_label_area_size(30)
      .y_label_area_size(30)
      .build_cartesian_2d(-1_f64..1_f64, -1_f64..1_f64)?;
    chart.configure_mesh().x_labels(20).y_labels(20).draw()?;
    Ok(())
  }
  ```

### DIFF
--- a/plotters-backend/src/lib.rs
+++ b/plotters-backend/src/lib.rs
@@ -105,7 +105,7 @@ impl<E: Error + Send + Sync> Error for DrawingErrorKind<E> {}
 ///  will use the pixel-based approach to draw other types of low-level shapes.
 pub trait DrawingBackend: Sized {
     /// The error type reported by the backend
-    type ErrorType: Error + Send + Sync;
+    type ErrorType: Error + Send + Sync + 'static;
 
     /// Get the dimension of the drawing backend in pixels
     fn get_size(&self) -> (u32, u32);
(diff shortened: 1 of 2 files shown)


AaronErhardt:
Currently, the master branch is used for the current stable series. Since this is a breaking change, you should target the next-release-devel branch. 

@38 maybe we should change that. I'd prefer having a stable branch and developing the upcoming version on master. That's also probably what most people expect.

raldone01:
@AaronErhardt should be ok now.

If I may chip in. Why not just develop on master/main and tag releases. If a stable release needs some bugfixes just branch at the tag and do the fixes.

AaronErhardt (approved):
You're right, I'd prefer having the development version on master. It makes more sense and the current branch situation has already caused quite a bit of confusion. But I'm not able to make this decision alone and the other maintainers are not very active atm.

PR #449, opened 2023-02-15
Easier interface for transforming 3d coordinate system

iverks:
This is not so much a real pr as it is a proof of concept, i believe i might have introduced breaking changes by not making Copy derivable.

I implemented an easier way of doing multiple transforms. The issue i had was that i wanted the z-axis to point toward the top of the screen, and i couldnt figure out how to with the current interface. This way to do it made it very simple. Example code below:

```rust
let mut chart =
            ChartBuilder::on(&root).build_cartesian_3d(-1.0..1.0, -1.0..1.0, -1.0..1.0)?;

chart.with_projection(|mut p| {
    p.add_transform(ProjectionMatrix::rotate(-PI * 0.5, 0.0, 0.0));
    p.scale = 0.9;
    p.into_matrix()
});
```

In context:

```rust
use plotters::prelude::*;
fn pdf(x: f64, y: f64) -> f64 {
    const SDX: f64 = 0.1;
    const SDY: f64 = 0.1;
    const A: f64 = 5.0;
    let x = x as f64 / 10.0;
    let y = y as f64 / 10.0;
    A * (-x * x / 2.0 / SDX / SDX - y * y / 2.0 / SDY / SDY).exp()
}

const OUT_FILE_NAME: &'static str = "plotters-doc-data/3d-plot2.gif";
    let root = BitMapBackend::gif(OUT_FILE_NAME, (600, 400), 100)?.into_drawing_area();

    for pitch in 0..157 {
        root.fill(&WHITE)?;

        let mut chart = ChartBuilder::on(&root)
            .caption("2D Gaussian PDF", ("sans-serif", 20))
            .build_cartesian_3d(-3.0..3.0, 0.0..6.0, -3.0..3.0)?;
        chart.with_projection(|mut p| {
            p.add_transform(ProjectionMatrix::rotate(-PI * 0.5 + , 0.0, 0.0)); // HERE
            p.pitch = 1.57 - (1.57 - pitch as f64 / 50.0).abs();
            p.scale = 0.7;
            p.into_matrix() // build the projection matrix
        });

        chart
            .configure_axes()
            .light_grid_style(BLACK.mix(0.15))
            .max_light_lines(3)
            .draw()?;

        chart.draw_series(
            SurfaceSeries::xoz(
                (-15..=15).map(|x| x as f64 / 5.0),
                (-15..=15).map(|x| x as f64 / 5.0),
                pdf,
            )
            .style_func(&|&v| {
                (&HSLColor(240.0 / 360.0 - 240.0 / 360.0 * v / 5.0, 1.0, 0.7)).into()
            }),
        )?;

        root.present()?;
    }

    // To avoid the IO failure being ignored silently, we manually call the present function
    root.present().expect("Unable to write result to file, please make sure 'plotters-doc-data' dir exists under current dir");
    println!("Result has been saved to {}", OUT_FILE_NAME);

    Ok(())
}
#[test]
fn entry_point() {
    main().unwrap()
}
```

Note that all projection matrices are applied before pitch and yaw from current interface, which might be confusing. 

I am willing to put some hours into fixing this feature if i get some good feedback on how to do it in a better way :)


AaronErhardt (changes requested):
Looks good, but since this is a breaking change, you should target the "next-release-devel" branch. Maybe we can get back to master branch eventually, but for now the master branch is for the current stable series.

iverks:
Rebasing

PR #452, opened 2023-02-25
Introduce Colormaps

jonaspleyer:
This PR aims to add Colormaps to plotters. It is only a first version with almost no documentation.
I am happy to discuss the actual implementations and traits and write some documentation when the design is finalized.
To sum up my efforts:
- Define a trait for a colormap
    - Take a (generic) floating point value (between 0.0 and 1.0) and return a corresponding color
    - Can also be called by normalizing with a provided minimum and maximum value
- Define macros to create a colormap from colorvalues or existing colors by linearly interpolating between the individual colors
    - This trait defines individual structs which names are the names of the colormap (ie. `Viridis {}`)
    - These structs can be called by their implemented trait functionality after they have been instantiated
    - or by associated functions (which currently do not belong to the trait but this could be changed).
- I have added a new feature "colormaps" and added it to the default features to optionally deactivate this functionality.
  This is of course debateable and could also be seen as opt-in rather than opt-out.
- I have modified the examples for 3d-plot2 and mandelbrot to showcase the new colormaps
- In addition I have added a new example showcasing currently existing colormaps
- There are still many opportunities to define new and beautiful colormaps

Feel free to critique me in my approach. It was simply a feature I needed myself while using the library which I would like to see added. Thanks for writing this very nice piece of software!

### DIFF
--- a/plotters/src/style/colors/colormaps.rs
+++ b/plotters/src/style/colors/colormaps.rs
@@ -1,3 +1,7 @@
+use crate::style::{HSLColor,RGBAColor,RGBColor};
+
+use num_traits::{Float,ToPrimitive,FromPrimitive};
+
 pub trait ColorScale<ColorType: crate::prelude::Color, FloatType=f32>
 where
     FloatType: Float,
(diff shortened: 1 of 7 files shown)


plotters/src/style/colors/colormaps.rs
```diff
@@ -0,0 +1,252 @@
...
+                let t = (h - min)/(max-min);
+                let approximate_index = t * (FloatType::from_usize(Self::COLORS.len()).unwrap() - FloatType::one()).max(FloatType::zero());
+                // Calculate which index are the two most nearest of the supplied value
+                let index_lower = approximate_index.floor().to_usize().unwrap();
+                let index_upper = approximate_index.ceil().to_usize().unwrap();
+                // Calculate the relative difference, ie. is the actual value more towards the color of index_upper or index_lower?
+                let relative_difference = approximate_index.ceil() - approximate_index;
+                // Interpolate the final color linearly
+                calculate_new_color_value!(relative_difference, Self::COLORS, index_upper, index_lower, $color_type)
+            }
```
AaronErhardt, line 141: There's a lot of duplicated code shared with `implement_color_scale_for_derived_ColorMap` that should probably be moved into a function.

plotters/src/style/colors/colormaps.rs
```diff
@@ -0,0 +1,252 @@
...
+        [$($color_type($($color_value),+)),+]
+    };
+    ($($color_complete:tt),+) => {
+        [$($color_complete),+]
+    };
+}
+
+
+macro_rules! implement_linear_interpolation_color_map{
+    ($color_scale_name:ident, $color_type:tt) => {
```
AaronErhardt, line 124:
```suggestion
    ($color_scale_name:ident, $color_type:ty) => {
```

plotters/src/style/colors/colormaps.rs
```diff
@@ -0,0 +1,252 @@
...
+            //           (1.0  - r)                   *                                           color_value_1  +                   r *                                           color_value_2
+            ((FloatType::one() - $relative_difference) * FloatType::from_f64($colors[$index_upper].0).unwrap() + $relative_difference * FloatType::from_f64($colors[$index_lower].0).unwrap()).to_f64().unwrap(),
+            ((FloatType::one() - $relative_difference) * FloatType::from_f64($colors[$index_upper].1).unwrap() + $relative_difference * FloatType::from_f64($colors[$index_lower].1).unwrap()).to_f64().unwrap(),
+            ((FloatType::one() - $relative_difference) * FloatType::from_f64($colors[$index_upper].2).unwrap() + $relative_difference * FloatType::from_f64($colors[$index_lower].2).unwrap()).to_f64().unwrap(),
+        )
+    };
+);
+
+
+macro_rules! implement_color_scale_for_derived_ColorMap{
```
AaronErhardt, line 79: This macro doesn't follow the naming conventions and should be called `implement_color_scale_for_derived_color_map` instead.

AaronErhardt (changes requested):
+ Prefer functions over macros due to better readability and cleaner scope. Your code uses a lot of macros, which makes sense in most places, but where it is possible, functions should be used instead (e.g. in the new colormaps example).
+ Use cargo fmt to format the code consistently. You can use `#[rustfmt::skip]` for code you want to keep formatted manually.
+ Replace `tt` (token tree which could be anything) with `ty` (which can only be a type). I already commented on some occurrences, but this has been done wrong in most macros.

jonaspleyer:
Thank you for taking the time to review my code! I have a few deadlines coming up at the moment but hope to be getting back to you afterwards. That will probably be in about 1-2 weeks.

plotters/src/style/colors/colormaps.rs
```diff
@@ -0,0 +1,252 @@
...
+    ($($color_type:tt),+) => {
```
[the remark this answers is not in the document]
jonaspleyer: I have corrected it :+1: It does not appear to be working with `ty` but rather with `ident`. This is probably due to how I match the supplied patterns. eg these lines:
```
($relative_difference:expr, $colors:expr, $index_upper:expr, $index_lower:expr, HSLColor) => {
```
I could not manage to make it work by actually accepting a type instead of an identifier but am not sure why.

plotters/src/style/colors/colormaps.rs
```diff
@@ -0,0 +1,252 @@
...
+        $(
+            impl<FloatType: Float + FromPrimitive + ToPrimitive> ColorMap<$color_type, FloatType> for DerivedColorMap<$color_type> {
+                fn get_color_normalized(&self, h: FloatType, min: FloatType, max: FloatType) -> $color_type {
+                    // Ensure that we do have a value in bounds
+                    let h = h.max(min).min(max);
+                    // Make sure that we really have a minimal value which is smaller than the maximal value
+                    assert_eq!(min<max, true);
```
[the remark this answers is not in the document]
jonaspleyer: Yes this makes more sense. Looks like some debug assert statement that I forgot.

plotters/src/style/colors/colormaps.rs
jonaspleyer: I have encapsulated the functionality into a function just as you described.

plotters/src/style/colors/colormaps.rs
jonaspleyer: See my comment above :smiley:

plotters/src/style/colors/colormaps.rs
jonaspleyer: was renamed :+1:

jonaspleyer:
I have addressed the concerns you had and have added documentation to the individual components. Please let me know if you still find that something is missing.

jonaspleyer:
Sorry for not noticing. Probably forgot to check again. Should be resolved now.

wangjiawen2013:
@jonaspleyer 
Hi,
Could you present a tutorial on how to use colormap ?  Can I add a colorbar to a plot using colormap ?

jonaspleyer:
I have not implemented colorbars next to the plots. Currently their functionality is rather limited. Some of the examples already use said colormaps. Have a look eg. at the Mandelbrot example. It has been some time since I have dealt with this implementation so I would need to take a closer look in order to really decide how to best proceed on this topic. Unfortunately I have many other obligations at the moment.

Do you have any specific use-case in mind or is this a general question?

wangjiawen2013:
colorbar is used frequently in my work. I add it to my plots to show the continuous data.

jonaspleyer:
I have to be honest: I cannot justify spending any time on this request currently. I might revisit this in the future however.

Issue #463, opened 2023-04-06
[BUG] TextStyle::pos is not used in MeshStyle::*_label_style


VincentFoulon80:
**Describe the bug**
When setting anchors on the label style for axes, the anchors are not used and arbitrary values are set instead.

This is particularly annoying when tilting the text to fit them in the chart (example modifying area chart demo):
![area-chart](https://user-images.githubusercontent.com/26797476/230505656-a56238ae-28fc-4ce6-9b9e-ad825a977af9.png)

**To Reproduce**
Here's a diff you can apply to the project, changing `area-chart` example to create the issue. The result is identical to the image above.
```diff
diff --git a/plotters/examples/area-chart.rs b/plotters/examples/area-chart.rs
index 5b1a7a5..79280c3 100644
--- a/plotters/examples/area-chart.rs
+++ b/plotters/examples/area-chart.rs
@@ -1,5 +1,6 @@
 use plotters::prelude::*;

+use plotters_backend::text_anchor::{HPos, Pos, VPos};
 use rand::SeedableRng;
 use rand_distr::{Distribution, Normal};
 use rand_xorshift::XorShiftRng;
@@ -28,10 +29,17 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
         .caption("Area Chart Demo", ("sans-serif", 40))
         .build_cartesian_2d(0..(data.len() - 1), 0.0..1500.0)?;

+    let label_style = BLACK
+        .with_anchor::<RGBColor>(Pos::new(HPos::Right, VPos::Center))
+        .into_text_style(&root)
+        .transform(FontTransform::Rotate270);
+
     chart
         .configure_mesh()
         .disable_x_mesh()
         .disable_y_mesh()
+        .x_label_formatter(&|x| String::from("HELLOWORLD"))
+        .x_label_style(label_style)
         .draw()?;

     chart.draw_series(
```

**Version Information**
Plotters 0.3.4

VincentFoulon80:
Fixed by #464

PR #464, opened 2023-04-06
Add HPos and VPos::Default to allow the user to override axes label anchors

VincentFoulon80:
Proposition for #463

Add `HPos::Default` and `VPos::Default` variants, that allows to detect intentional changes to anchors and so prevent using unexpected change of position (The documentation didn't tell us axes overrides positioning values).

I decided to create new variants to address the issue of detecting that the user has changed the anchors of one axis's labels. (because the default Pos are `HPos::Left` and `VPos::Top`)

One other possibility is to change how the position is decided in `draw_impl`, based on the `FontTransform`. This may solve the issue as is but still won't allow customization of the position, and additionally we'd want to change the documentation in order to tell the user that positions are not used for axis labels.

### DIFF
--- a/plotters-backend/src/text.rs
+++ b/plotters-backend/src/text.rs
@@ -60,10 +60,11 @@ impl<'a> From<&'a str> for FontFamily<'a> {
 /// ```
 pub mod text_anchor {
     /// The horizontal position of the anchor point relative to the text.
-    #[derive(Clone, Copy)]
+    #[derive(Default, Clone, Copy)]
     pub enum HPos {
         /// Default value (Left), except chart axes that might provide a different pos.
         /// Use another variant to override the default positionning
+        #[default]
         Default,
         /// Anchor point is on the left side of the text
         Left,
@@ -74,10 +75,11 @@ pub mod text_anchor {
     }
 
     /// The vertical position of the anchor point relative to the text.
-    #[derive(Clone, Copy)]
+    #[derive(Default, Clone, Copy)]
     pub enum VPos {
         /// Default value (Top), except chart axes that might provide a different pos
         /// Use another variant to override the default positionning
+        #[default]
         Default,
         /// Anchor point is on the top of the text
         Top,
@@ -88,7 +90,7 @@ pub mod text_anchor {
     }
 
     /// The text anchor position.
-    #[derive(Clone, Copy)]
+    #[derive(Default, Clone, Copy)]
     pub struct Pos {
         /// The horizontal position of the anchor point
         pub h_pos: HPos,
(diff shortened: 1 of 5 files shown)


plotters-backend/src/text.rs
```diff
@@ -62,6 +62,9 @@ pub mod text_anchor {
     /// The horizontal position of the anchor point relative to the text.
     #[derive(Clone, Copy)]
     pub enum HPos {
+        /// Default value (Left), except chart axes that might provide a different pos.
+        /// Use another variant to override the default positionning
+        Default,
```
AaronErhardt, line 67: I think it makes sense to derive `Default` for both HPos and VPos since they already have a `Default` variant.

AaronErhardt (changes requested):
Looks good in general.

plotters-backend/src/text.rs
VincentFoulon80: I am not sure I understood what you meant. Did you mean something like this ?
```diff
diff --git a/plotters-backend/src/text.rs b/plotters-backend/src/text.rs
index bcf3a7a..4acd931 100644
--- a/plotters-backend/src/text.rs
+++ b/plotters-backend/src/text.rs
@@ -60,10 +60,11 @@ impl<'a> From<&'a str> for FontFamily<'a> {
 /// ```
 pub mod text_anchor {
     /// The horizontal position of the anchor point relative to the text.
-    #[derive(Clone, Copy)]
+    #[derive(Default, Clone, Copy)]
     pub enum HPos {
         /// Default value (Left), except chart axes that might provide a different pos.
         /// Use another variant to override the default positionning
+        #[default]
         Default,
         /// Anchor point is on the left side of the text
         Left,
@@ -74,10 +75,11 @@ pub mod text_anchor {
     }

     /// The vertical position of the anchor point relative to the text.
-    #[derive(Clone, Copy)]
+    #[derive(Default, Clone, Copy)]
     pub enum VPos {
         /// Default value (Top), except chart axes that might provide a different pos
         /// Use another variant to override the default positionning
+        #[default]
         Default,
         /// Anchor point is on the top of the text
         Top,
@@ -88,7 +90,7 @@ pub mod text_anchor {
     }

     /// The text anchor position.
-    #[derive(Clone, Copy)]
+    #[derive(Default, Clone, Copy)]
     pub struct Pos {
         /// The horizontal position of the anchor point
         pub h_pos: HPos,
@@ -111,22 +113,6 @@ pub mod text_anchor {
         pub fn new(h_pos: HPos, v_pos: VPos) -> Self {
             Pos { h_pos, v_pos }
         }
-
-        /// Create a default text anchor position (top left).
-        ///
-        /// - **returns** The default text anchor position
-        ///
-        /// ```rust
-        /// use plotters_backend::text_anchor::{Pos, HPos, VPos};
-        ///
-        /// let pos = Pos::default();
-        /// ```
-        pub fn default() -> Self {
-            Pos {
-                h_pos: HPos::Default,
-                v_pos: VPos::Default,
-            }
-        }
     }
 }
```

AaronErhardt (approved):
Looks good, thanks!

cszach:
Was this change included in 0.3.5? I used `HPos::Right, VPos::Center` like the code in #463 but still got the result like in the example plot.

AaronErhardt:
No, this was not included because is was merged into `next-release-devel` instead of master. Historically, master is considered the stable branch and `next-release-devel` is for the next breaking semver release. I'm sorry if this caused confusion, I hope we can clean this up soon and switch to better branch names.

cszach:
Thank you for the clarification! So at the moment how do I align text properly? Some text are getting into the graphing area and I want them to align nicely below the axis. Thanks in advance!

image

AaronErhardt:
I think the alignment is still broken both on `master` and crates.io, so you need to use the `next-release-devel` branch. To make this work properly, this fix would need to be backported to the master branch and then we can publish a new release.

PR #467, opened 2023-04-20
Fix unaligned read/write in bitmap backend.

### LINKED ISSUE #461: [BUG] Panic in unsafe `fill_rect_fast`
**Describe the bug**

When using Rust nightly compiler, attempting to draw a line is likely to cause a panic (depending on color / line coordinates):

```
thread 'main' panicked at 'misaligned pointer dereference: address must be a multiple of 0x8 but is 0x1ab590c7fd3', [redacted]\plotters-bitmap-0.3.2\src\bitmap_pixel\rgb.rs:215:29
```

The offending line is: https://github.com/plotters-rs/plotters/blob/acf5158c6f49b8bfe8a5431c64b0867a1a8b36d1/plotters-bitmap/src/bitmap_pixel/rgb.rs#L215

I believe the relevant Nightly change is https://github.com/rust-lang/rust/pull/98112, but this seems to be a plotters bug and not a rustc regression (from my limited understanding).

**To Reproduce**

```rust
use plotters::prelude::*;

fn main() {
    let mut backend = BitMapBackend::new("test.png", (256, 256));
    backend.draw_line((1, 0), (100, 0), &RED).unwrap();
}
```
Sample project: https://github.com/agersant/plotter-panic

**Version Information**
Plotters `0.3.4`

Berrysoft:
Fixes #461

### DIFF
--- a/plotters-bitmap/src/bitmap_pixel/bgrx.rs
+++ b/plotters-bitmap/src/bitmap_pixel/bgrx.rs
@@ -87,7 +87,7 @@ impl PixelFormat for BGRXPixel {
             let slice = unsafe { std::slice::from_raw_parts_mut(start_ptr, (count - 1) / 2) };
             for rp in slice.iter_mut() {
                 let ptr = rp as *mut [u8; 8] as *mut u64;
-                let d1 = unsafe { *ptr };
+                let d1 = unsafe { ptr.read_unaligned() };
                 let mut h = (d1 >> 8) & M;
                 let mut l = d1 & M;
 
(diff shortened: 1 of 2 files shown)


nbdd0121:
Please fix the reliance on tuple layout as well (i.e. change `(u64, u64, u64)` to `[u64; 3]`)

AaronErhardt (approved):
Thanks! Could you check whether the `next-release-devel` branch that's used for the upcoming version is also affected?

Berrysoft:
@AaronErhardt yes, it is also affected.

AaronErhardt:
Would you mind sending a PR for that branch as well?

Berrysoft:
I've opened a PR: #469

PR #480, opened 2023-06-17
delete whitespace so cargo doc will format comment in list

zancas:
Make this doc-comment should be rendered as list when processed by `cargo doc`.


AaronErhardt:
Thanks!

PR #481, opened 2023-06-27
Enable macro usage for users to define colormaps

jonaspleyer:
This PR enables the user to leverage the macro `define_linear_interpolation_color_map` to easily create custom colormaps.

In addition, I have hidden the documentation of macros in the crate root and instead included them in the correct module.

Some functionalities which are required are hidden since they are not meant to be used by the end-user.

### DIFF
--- a/plotters/blub.png
+++ b/plotters/blub.png
@@ -1,0 +1,3 @@
+<svg width="800" height="600" viewBox="0 0 800 600" xmlns="http://www.w3.org/2000/svg">
+<line opacity="1" stroke="#000000" stroke-width="0" x1="338" y1="122" x2="365" y2="122"/>
+</svg>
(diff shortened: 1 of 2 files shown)


jonaspleyer:
It seems that one of the checks was canceled. I cannot see any apparent reason as to why I would have happened.

AaronErhardt:
I've re-run the canceled check. I've no idea why it was canceled either.

Anyway, I wonder whether it makes sense to use `#[doc(hidden)]` followed by doc comments. AFAIK there will be no docs generated for hidden items so using doc comments on such items doesn't really make sense, or am I wrong? Can you maybe show how the docs look for you locally?

jonaspleyer:
The documentation is hidden on my machine. To me it makes sense if another developer is looking at your code and trying to understand it. Maybe I was too motivated in writing doc-tests. Since they present an actual computational overhead, it might be reasonable to omit them.

jonaspleyer:
To advance this PR: I moved the relevant docstrings to be inline comments only. Furthermore, I hid the `count!` macro since it is only being used as a helper macro and shortened the macro to define colormaps in order to be easier to use and not having to write such a long name.

Please let me know if this fixes your previously mentioned concern.

AaronErhardt:
The code looks good, but I think there are a couple of plots that should not be part of the commits.

jonaspleyer:
Sorry this was a total oversight on my part. I have rebased my master branch to not include the plots in the git history at all.

AaronErhardt (approved):
Looks good, thanks!

PR #485, opened 2023-07-16
Feat: added method to Pie Chart for creating a Donut Chart

GyulyVGC:
This PR adds a new method `donut_hole` to the Pie element, making it possible to create Donut Charts with a specified hole radius.

If a value minor than zero or major than the Pie radius is passed for the donut hole, it will be ignored.

The percentages labels are put in the middle of the donut area.

image

### DIFF
--- a/plotters/src/element/pie.rs
+++ b/plotters/src/element/pie.rs
@@ -189,11 +189,7 @@ impl<'a, DB: DrawingBackend, Label: Display> Drawable<DB> for Pie<'a, (i32, i32)
                 let label_size = backend.estimate_text_size(&perc_label, percentage_style)?;
                 let text_x_mid = (label_size.0 as f64 / 2.0).round() as i32;
                 let text_y_mid = (label_size.1 as f64 / 2.0).round() as i32;
-                let perc_radius = if self.donut_hole == 0.0 {
-                    self.radius / 2.0
-                } else {
-                    (self.radius + self.donut_hole) / 2.0
-                };
+                let perc_radius = (self.radius + self.donut_hole) / 2.0;
                 let perc_coord = theta_to_ordinal_coord(
                     perc_radius,
                     middle_theta,
--- a/plotters/src/element/pie.rs
+++ b/plotters/src/element/pie.rs
@@ -31,7 +31,7 @@ pub struct Pie<'a, Coord, Label: Display> {
     label_style: TextStyle<'a>,
     label_offset: f64,
     percentage_style: Option<TextStyle<'a>>,
-    donut_hole: &'a f64, // radius of the hole in case of a donut chart
+    donut_hole: f64, // radius of the hole in case of a donut chart
 }
 
 impl<'a, Label: Display> Pie<'a, (i32, i32), Label> {
@@ -63,7 +63,7 @@ impl<'a, Label: Display> Pie<'a, (i32, i32), Label> {
             label_style,
             label_offset: radius_5pct,
             percentage_style: None,
-            donut_hole: &0.0,
+            donut_hole: 0.0,
         }
     }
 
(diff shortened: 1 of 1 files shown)


plotters/src/element/pie.rs
```diff
@@ -135,9 +152,18 @@ impl<'a, DB: DrawingBackend, Label: Display> Drawable<DB> for Pie<'a, (i32, i32)
...
+                    offset_theta -= radian_increment;
+                }
+            }
             // final point of the wedge may not fall exactly on a radian, so add it extra
-            let final_coord = theta_to_ordinal_coord(*self.radius, theta_final, self.center);
-            points.push(final_coord);
+            if self.donut_hole == &0.0 {
+                let final_coord = theta_to_ordinal_coord(*self.radius, theta_final, self.center);
+                points.push(final_coord);
+            }
```
AaronErhardt, line 166: I'm not entirely sure whether that's the best approach. The idea behind the final point was to make it possible to hit the position exactly, even if no increment of `radian_increment` fits in the remaining space. I would keep this unconditional and would start subtracting from `final_coord` on the way back and then calculate a second `final_coord` for the final inner edge.

plotters/src/element/pie.rs
```diff
@@ -31,6 +31,7 @@ pub struct Pie<'a, Coord, Label: Display> {
     label_style: TextStyle<'a>,
     label_offset: f64,
     percentage_style: Option<TextStyle<'a>>,
+    donut_hole: &'a f64, // radius of the hole in case of a donut chart
```
AaronErhardt, line 34: Is there any reason for a lifetime here?

plotters/src/element/pie.rs
```diff
@@ -163,8 +189,13 @@ impl<'a, DB: DrawingBackend, Label: Display> Drawable<DB> for Pie<'a, (i32, i32)
                 let label_size = backend.estimate_text_size(&perc_label, percentage_style)?;
                 let text_x_mid = (label_size.0 as f64 / 2.0).round() as i32;
                 let text_y_mid = (label_size.1 as f64 / 2.0).round() as i32;
+                let perc_radius = if self.donut_hole == &0.0 {
+                    self.radius / 2.0
+                } else {
+                    (self.radius + self.donut_hole) / 2.0
```
AaronErhardt, line 195: I think the equation is wrong. 
If written like this, you don't even need the new the if statement because if `self.donut_hole == &0.0` is true, `(self.radius + self.donut_hole) / 2.0` is identical to `self.radius / 2.0`.

I think the correct equation for the donut case would be `self.radius + self.donut_hole / 2.0`.

AaronErhardt (changes requested):
The new feature looks really nice, but there are a couple of things I would like to have addressed before merging.

plotters/src/element/pie.rs
GyulyVGC: Actually nope. Fixed by https://github.com/plotters-rs/plotters/pull/485/commits/fbfefd6f2c01c2077a06a8cace249f7ddf4cd6bf

plotters/src/element/pie.rs
GyulyVGC: The correct equation is that originally written by me. We need the average between the external radius and the internal one.
But as you noticed, actually there is no need of the if statement.

GyulyVGC: Fixed in https://github.com/plotters-rs/plotters/pull/485/commits/8f39a5eda0fee945c24810f10d0cecb6e7055fd6

GyulyVGC:
I'm done with the adjustments. Let me know!

AaronErhardt (approved):
Thanks, looks good.

PR #508, opened 2023-10-17
Chore: "Cargo doc" removed this warning.

martinfrances107:
```
warning: unclosed HTML tag `u8`
  --> plotters-bitmap/src/bitmap.rs:94:95
   |
94 |     /// When this is used, the bitmap backend will write to a user provided [u8] array (or Vec<u8>)
   |                                                                                               ^^^^
   |
   = note: `#[warn(rustdoc::invalid_html_tags)]` on by default
help: try marking as source code
   |
94 |     /// When this is used, the bitmap backend will write to a user provided [u8] array (or `Vec<u8>`)
```


AaronErhardt:
Thanks!

PR #509, opened 2023-10-17
Chore: ran cargo clippy --fix

martinfrances107:
Also ran cargo fmt.

After inspecting all the changes,

There is one case where we can remove a .clone() call, but it is not in a hot path, so the significance is low.

```
-        let end = start.clone() + Duration::nanoseconds(100);
+        let end = start + Duration::nanoseconds(100);
```

There a lots of cases where we can remove a cast for example.

```
-                total_ns as u64,
+                total_ns,
```


AaronErhardt (approved):
Thanks, looks good.

PR #511, opened 2023-10-17
BugFix: Error propagation is silenced

martinfrances107:
```
warning: unused `Result` that must be used
   --> plotters/src/style/font/ttf.rs:135:9
    |
135 | /         data.clone().map(|handle| {
136 | |             handle
137 | |                 .load()
138 | |                 .map(FontExt::new)
139 | |                 .map_err(|e| FontError::FontLoadError(Arc::new(e)))
140 | |         })?;
    | |___________^
    |
    = note: this `Result` may be an `Err` variant, which should be handled
    = note: `#[warn(unused_must_use)]` on by default
help: use `let _ = ...` to ignore the resulting value
    |
135 |         let _ = data.clone().map(|handle| {
```

When I remove the ? and look at the type of the droppped variable I see that is it Result<Result<FontExt, FontError>, FontError>

Rust is very unintuative on this matter. The solution is to double unwrap.

There are many ways to do this .. I went with a soltions I think is readable.

### DIFF
--- a/plotters/src/style/font/ttf.rs
+++ b/plotters/src/style/font/ttf.rs
@@ -132,16 +132,13 @@ fn load_font_data(face: FontFamily, style: FontStyle) -> FontResult<FontExt> {
     // Then we need to check if the data cache contains the font data
     let cache = DATA_CACHE.read().unwrap();
     if let Some(data) = cache.get(Borrow::<str>::borrow(&key)) {
-        let font_ext = data.clone().map(|handle| {
+        data.clone().map(|handle| {
             handle
                 .load()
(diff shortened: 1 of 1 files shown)


AaronErhardt:
Does a double question mark `??` or flatten with `.flatten()?` not work in this case?

martinfrances107:
I did not event consider using ??.. Readability is such a subjective thing.

I am happy enough to get behind this as a simplification.... 

and single character bugfixes make me happy.

AaronErhardt (approved):
Thanks!

Issue #530, opened 2023-12-28
[BUG] types/datetime: replace deprecated chrono::Date* with NaiveDate


10ne1:
Hello

**Describe the bug**
Starting with chrono 0.4.23, the Date, DateTime and *_ymd() APIs got deprecated and will eventually be removed, so we should update the code to use the new APIs. Fortunately support for NaiveDate is already implemented so we just need to update some function calls and remove the deprecated trait implementations.

**To Reproduce**
Build plotters with chrono >= 0.4.23 and notice the deprecation warnings. Tested both normal target builds and  wasm32-unknown-unknown.

**Version Information**
I use the git mainline branch, for which I created a fix which I'll post shortly closing this issue. Thanks!

PR #531, opened 2023-12-28
types/datetime: replace deprecated chrono::Date* with NaiveDate

10ne1:
The chrono crate deprecated the Date and DateTime APIs in favor of NaiveDate* and also added interfaces like from_ymd_opt() which return Options.

Eventually the deprecated APIs will be removed so this migrates to the new APIs (NaiveDate is already implemented, so we just do the minimal required migration).

Deprecation  started since 0.4.23:
https://docs.rs/chrono/0.4.31/chrono/struct.Date.html

While at it, also bump the chrono dep 0.4.20 -> 0.4.31

Closes: #530

### DIFF
--- a/plotters/Cargo.toml
+++ b/plotters/Cargo.toml
@@ -14,7 +14,7 @@ exclude = ["doc-template", "plotters-doc-data"]
 
 [dependencies]
 num-traits = "0.2.14"
-chrono = { version = "0.4.19", optional = true }
+chrono = { version = "0.4.31", optional = true }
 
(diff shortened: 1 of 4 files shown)


AaronErhardt:
It seems to me that this is a breaking change. Contrary to most other projects, plotters uses "master" as stable branch and "next-release-devel" for semver breaking releases. Would you mind targeting the "next-release-devel" branch with this PR?

AaronErhardt (approved):
Looks good. Thanks!

PR #483, opened 2023-07-01
Add dashed line style.

KmolYuan:
A simple work similar to `LineSeries` and `PathElement`.

# New Types

Added `DashedLineSeries` (under `line_series` feature) and `DashedPathElement`.

+ `plotters::series::line_series::DashedLineSeries` - A simple type just making `std::iter::once(element)` by implementing `IntoIterator`, and has the same options as the element type.
+ `plotters::element::basic_shapes::DashedPathElement` - A element type drawing dashed lines pixel-wised.

# Options

+ `points`: The iterator of the points.
+ `size`: The dash size.
+ `spacing`: The dash-to-dash spacing (gap size).
+ `style`: The shape style.

# ⚠️ Known Issue

+ ~If the points are too dense, it will become a solid line.~ (solved)
+ Distortion in rounding coordinates, mainly in SVG.

# Preview

Use blue line. The cross markers are the input points.

```rust
let series = DashedLineSeries::new(points, 10, 5, color);
chart.draw_series(series)?.label(label).legend(move |(x, y)| {
    DashedPathElement::new([(x, y), (x + 20, y)], 10, 5, color)
});
```

Another SVG example of my work is 3D coordinates. The curves are projected on a sphere, the main shape using a red dashed line, and the mechanism behind the sphere is shown with gray dashed lines.

PR #540, opened 2024-01-20
Fix dashed line algorithm

KmolYuan:
#483
I'm sorry, I found some problems using the dashed line algorithm. When the distance between points is greater than the `size` or `spacing` of the dashed line, the algorithm will produce an incorrect continuous straight line.

This new algorithm was tested using two curve coordinate sequences (as shown in the following), **"point distance less than `size`"** and **"point distance greater than `size`,"** the above results were corrected.

In addition, if the `size` equals 0 pixel, it will not be drawn.

**Left:** point distance less than `size` or `spacing`
**Right:** point distance greater than `size` or `spacing`
<div>

</div>


AaronErhardt (approved):
Looks good, thanks!

Issue #551, opened 2024-02-20
[BUG] Rotated labels overlap with the chart


SimonBrandner:
**Describe the bug**

When drawing rotated axis labels (e.g. dates and times), they should not overlap with the chart -- their origin should not be at the center but at the top

![image](https://user-images.githubusercontent.com/228850/233118314-1b6f88e2-041c-455c-9f09-a56869bd037f.png)

**Version Information**

`0.3.5`

**Additional info**

See https://github.com/orgs/plotters-rs/discussions/466

Pierstoval:
I have the same problem. I tried using `.pos(Pos::new(HPos::Left, VPos::Top))` alongside the font transform, but it didn't do anything 😕

PR #555, opened 2024-02-28
Fix [BUG] Rotated labels overlap with the chart #551

AliMMehr:
Do not override the `label_style.pos` value if it has been defined by the user.

This is the [related discussion](https://github.com/orgs/plotters-rs/discussions/466) for issue #551 . I also commented out a rational for the fix in [here](https://github.com/orgs/plotters-rs/discussions/466#discussioncomment-8589636)

This is a breaking change, but the only breaking change is that the type of `TextStyle.pos` changes from `text_anchor::Pos` to `Option<text_anchor::Pos>`.

### DIFF
--- a/plotters-backend/src/lib.rs
+++ b/plotters-backend/src/lib.rs
@@ -235,12 +235,14 @@ pub trait DrawingBackend: Sized {
         let ((min_x, min_y), (max_x, max_y)) = layout;
         let width = (max_x - min_x) as i32;
         let height = (max_y - min_y) as i32;
-        let dx = match style.anchor().h_pos {
+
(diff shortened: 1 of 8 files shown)


AaronErhardt:
> I don't know if this kind of breaking change is allowed. I also don't know if this is a good fix for this issue. Please advise any type of change you would want me to do, and I can do it. Thank you

The fix looks good to me, but breaking changes are not accepted on the `master` branch. For breaking changes, please submit a merge request to `next-release-devel`. Unfortunately however, due to lack of maintainers, there is no estimate currently when the next major version will be published.

PR #561, opened 2024-03-21
Clamp infinite values in logarithmic coordinate system

NorbertGarfield:
Solves https://github.com/plotters-rs/plotters/issues/532

When `LogCoord<V>::map` gets a value equal to zero, [fv.ln()](https://github.com/plotters-rs/plotters/blob/af0b63ca2c0ca18780ec9e92730fa97a444dbfca/plotters/src/coord/ranged1d/combinators/logarithmic.rs#L194) yields -∞.
Map positive and negative infinity to the upper and lower limits to avoid panicking.


NorbertGarfield:
@AaronErhardt, please review.

AaronErhardt (approved):
Thanks, looks good.

PR #571, opened 2024-04-14
Multiline text layout panics when multi-byte UTF-8 is used

oswald2:
The layout function for multi-line text crashes when multi-byte UTF-8 characters are used.  This pull request fixes this issue.

The old algorithm was simply operating on bytes and directly indexing the string slices on byte boundaries, which panics for UTF-8 multi-byte characters (one of my use-cases). 

The function has been rewritten to use an iterator over the string slice via `str::char_indices`, which is on base of a `char` and not of a byte and therefore avoids the panic. A small test case was added in the code. The test depends on the ttf feature though, as it must use a font to determine the width of a text in pixels. So, the test is only compiled and run when also the ttf feature is enabled. 

NOTE: the CONTRIBUTING file mentions that plotters should be compilable with Rust 1.36. Unfortunately, I didn't manage to do that. After setting the toolchain, it first complained about the ab_glyph feature having the same name as a dependency. Fixing this, the compilation ran into a problem as the newer version of chrono (which are automatically downloaded on the build with 1.36) are for edition 2021, which is then of course unrecognized as 1.36 is for edition 2018. After trying around a bit, I could not get this to work with 1.36, so I stopped. I assume, that plotters itself is normally compiled with newer versions than 1.36 as the ab_glyph feature and chrono would not allow that. I hope this is ok.


.gitignore
```diff
@@ -4,3 +4,21 @@
 .vscode/*
 Cargo.lock
 .idea
+plotters/area_series.svg
+plotters/chart_builder_on.svg
+plotters/composable.svg
+plotters/configure_axes.svg
```
jayvdb, line 10: Maybe change this list to a single entry `plotters/*.svg`  ?

oswald2: Oops, yes, thank you! 
Done.

AaronErhardt (approved):
Looks good, thanks!

PR #592, opened 2024-05-31
[FIX] fix datetime addition

holg:
The chrono::Duration object doesn't have += operation defined. So instead of using +=, the addition should be done explicitly.
It worked before, but this made it not being compiled, so i fixed it.

### DIFF
--- a/plotters/src/coord/ranged1d/types/datetime.rs
+++ b/plotters/src/coord/ranged1d/types/datetime.rs
@@ -779,7 +779,7 @@ impl Ranged for RangedDuration {
 
                 while current < self.1 {
                     ret.push(current);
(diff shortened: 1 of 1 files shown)


AaronErhardt:
It seems like support for `+=` was [added in chrono 0.4.34](https://docs.rs/chrono/0.4.34/chrono/type.Duration.html#impl-AddAssign-for-TimeDelta), but plotters only lists v0.4.20 as baseline requirement. So instead of undoing the recent commits that fixed some clippy lints, I suggest to update the baseline for chrono to 0.4.34.

holg:
Instead of doing this fix the update the baseline for chrono to 0.4.34.

PR #593, opened 2024-06-04
Clarify Quick Start procedure so program works on first try

Soccertanker:
Previously the quadratic function chart Quick Start instructions did not tell the user to create the subdir plotters-doc-data, where the example code writes 0.png.

When the subdir plotters-doc-data does not exist, the code appropriately panics. This does not leave the user with a good first impression of the repository & rust crate.

Now the instructions clearly tell the user to create the subdirectory before running the example program.


AaronErhardt (approved):
Thanks!

PR #596, opened 2024-06-09
Add fedora dependencies to readme

MrSauna:


MrSauna:
It would better to include the actual dependencies instead of platform specific packages. Platform specific dependencies are a bonus.

AaronErhardt (approved):
Thanks!

PR #599, opened 2024-06-22
pie chart supports RGBAColor

wangjiawen2013:
Make pie chart support RGBAColor, which is very usefull when overlaying multi-layers plots.

### DIFF
--- a/plotters/src/element/pie.rs
+++ b/plotters/src/element/pie.rs
@@ -71,7 +71,7 @@ impl<'a, Label: Display> Pie<'a, (i32, i32), Label> {
     /// Default is set to start at 0, which is aligned on the x axis.
     /// ```
     /// use plotters::prelude::*;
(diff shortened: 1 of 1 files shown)


wangjiawen2013:
This will solve the issue: https://github.com/plotters-rs/plotters/issues/597

AaronErhardt:
Since this is a breaking change, I cannot merge this into master but only `next-relase-devel`. Can you change the target branch?

wangjiawen2013:
Ok, I will change the target branch. By the way, what is "breaking change" ? Will you merge changes which are not "breaking change" in master (such as `evcxr_figure_with_saving`) to "next-release-devel" ？If not, There will be more and more difference between master and next-release-devel and the features/enhancements in master will not appear in next-release-devel.

wangjiawen2013:
Ok, I will change the target branch. By the way, what is "breaking change" ? Will you merge changes which are not "breaking change" in master (such as `evcxr_figure_with_saving`) to "next-release-devel" ？If not, There will be more and more difference between master and next-release-devel and the features/enhancements in master will not appear in next-release-devel.

AaronErhardt:
> By the way, what is "breaking change"?

https://doc.rust-lang.org/cargo/reference/semver.html

PR #600, opened 2024-06-23
pie chart supports RGBAColor

wangjiawen2013:
Make pie chart support RGBAColor, which is very usefull when overlaying multi-layers plots. This will solve the issue: https://github.com/plotters-rs/plotters/issues/597


wangjiawen2013:
Ok

PR #601, opened 2024-06-23
element/basic_shapes: add two Rectangle getters/setters

10ne1:
It is useful to be able to update rectangle styles after the rectangles were created, so we add a helper for it.

It is also useful to be able to get the point coordinates.


AaronErhardt (approved):
Looks good, thanks!

PR #602, opened 2024-06-23
style/font: use web font style

10ne1:
style/font: use web font style

The font-style is actually useful so let's set it.


AaronErhardt:
Since this appears to be an internal API I think we should rather fix the warning by removing the unnecessary field instead of suppressing the warning. Or is there a reason why we need the second field?

AaronErhardt (approved):
Thanks!

PR #610, opened 2024-07-07
Relax too strict life times on ChartBuilder

jchacks:
The `ChartBuilder`'s lifetimes meant that the `root_area` was not allowed to be dropped after the builder has been finalised (i.e. with a call to `build_cartesian_2d`).  I'm not 100% sure but this feels like unintended behaviour.

Since the `root_area` is cloned and no references to the original (usually `root` DrawingArea) are maintained by the resulting `ChartContext` this lifetime can be relaxed.

This should allow for `ChartContext`s to outlive the `root` drawing areas.

```rust
let drawing_area = create_mocked_drawing_area(640, 480, |_| {});
let mut chart = ChartBuilder::on(&drawing_area)
    .build_cartesian_2d(0f32..10f32, (1e-6f32..1f32).log_scale())
    .unwrap();
drop(drawing_area); // This drop would not be allowed before
// do something with `chart` 
```

### DIFF
--- a/plotters/src/chart/builder.rs
+++ b/plotters/src/chart/builder.rs
@@ -285,12 +285,12 @@ impl<'a, 'b, DB: DrawingBackend> ChartBuilder<'a, 'b, DB> {
     #[deprecated(
         note = "`build_ranged` has been renamed to `build_cartesian_2d` and is to be removed in the future."
     )]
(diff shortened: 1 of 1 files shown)


AaronErhardt (approved):
Thanks!

PR #612, opened 2024-07-11
RangedCoordisize made public

robertosw:
I think this has just been forgotten. I did not compile this locally only checked for similarities with `RangedCoordusize` and from searching through the repo it seemed like  `RangedCoordisize` was just as "widely" implemented and usable as its `usize`.


AaronErhardt (approved):
Thanks!

Issue #321, opened 2022-01-04
[Feature Request] Line style 


DaMilyutin:
### What is the feature ?
It would be nice to have not only solid lines but also dashed, dotted or whatever pattern.
I haven't found this in documentation. So, if this functionality exists, consider this issue a request for doc.

### (Optional) Why this feature is useful and how people would use the feature ?
There are lots of use cases when you need different line styles to distinguish plots.
Ex. when colors and width is not enough or is less expressive. 
This functionality exists in different languages (see links below) so obviously there are a lot of users for this.

### (Optional) Additional Information
Examples from other languages. 
Matlab: https://www.mathworks.com/help/matlab/creating_plots/specify-line-and-marker-appearance-in-plots.html
Scilab: https://help.scilab.org/docs/5.5.1/en_US/LineSpec.html
Python/matplotlib: https://matplotlib.org/2.0.0/api/lines_api.html 
C++/Qt/QCustomPlot: https://www.qcustomplot.com/index.php/tutorials/basicplotting

DaMilyutin:
I did some research on this. See results
https://github.com/plotters-rs/plotters/discussions/324

PR #616, opened 2024-07-16
Examples in Docs of DashedLineSeries

robertosw:
This might be what is wanted in #321 

This is such a small example that I didn't want to create a new file in `plotters/examples` for it. Is this okay?


robertosw:
btw: `ShapeStyle::filled` does not even have to be true. It will always get filled

AaronErhardt:
Thanks!

PR #625, opened 2024-08-03
plotters: update chrono dependency to 0.4.32

### LINKED ISSUE #624: [BUG] `plotters` fails to compile in a workspace using chrono 0.4.31
When adding a `plotters 0.3.6` dependency to an existing workspace, it fails to compile:

```
error[E0368]: binary assignment operation `+=` cannot be applied to type `chrono::Duration`
   --> .cargo/registry/src/index.crates.io-6f17d22bba15001f/plotters-0.3.6/src/coord/ranged1d/types/datetime.rs:782:21
    |
782 |                     current += Duration::nanoseconds(period as i64);
    |                     -------^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |                     |
    |                     cannot use `+=` on type `chrono::Duration`

error[E0368]: binary assignment operation `+=` cannot be applied to type `chrono::Duration`
   --> .cargo/registry/src/index.crates.io-6f17d22bba15001f/plotters-0.3.6/src/coord/ranged1d/types/datetime.rs:821:13
    |
821 |             current += Duration::days(i64::from(days_per_tick));
    |             -------^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |             |
    |             cannot use `+=` on type `chrono::Duration`

```
The cause is that the existing workspace already uses chrono 0.4.31, and the plotters crate does not correctly specify the version it requires.

The `plotters` crate declares a dependency on [chrono 0.4.20](https://github.com/plotters-rs/plotters/blob/dd1296adebda14bfc1d24e4a9c0309e334fa70f5/plotters/Cargo.toml#L17C1-L17C49), but actually uses features that only exist in a later version (chrono >= 0.4.34).

This was previously noticed in #592, but is still unfixed.

eric-seppanen:
The plotters code makes use of `AddAssign` on `chrono::Duration` (added in 24bf0638), but that wasn't added to `chrono` until 0.4.32.

Without this fix, if an existing workspace adds `plotters` but already has a `Cargo.lock` containing an earlier release of `chrono` 0.4.x, plotters will fail to compile.

Fixes #624.

### DIFF
--- a/plotters/Cargo.toml
+++ b/plotters/Cargo.toml
@@ -14,7 +14,7 @@ exclude = ["doc-template", "plotters-doc-data"]
 
 [dependencies]
 num-traits = "0.2.14"
-chrono = { version = "0.4.20", optional = true }
+chrono = { version = "0.4.32", optional = true }
 
 [dependencies.plotters-backend]
 version = "0.3.6"


AaronErhardt (approved):
Thanks!

PR #629, opened 2024-08-25
plotters: Depend on wasm-bindgen 0.2.89 or higher

workingjubilee:
wasm-bindgen 0.2.62 is not compatible with a wasm ABI change that rustc wishes to enable by default for wasm32-unknown-unknown, currently gated behind passing the -Zwasm-c-abi flag to rustc.

wasm-bindgen 0.2.89 should exhibit seamless behavior before and after the ABI change to match the C ABI, so depend on that.

For more information, see
- https://github.com/rust-lang/rust/issues/115666
- https://github.com/rust-lang/rust/pull/117918
- https://github.com/rust-lang/rust/issues/122532


AaronErhardt (approved):
Sounds like a reasonable change. Thanks!

PR #634, opened 2024-08-31
update svg backend code to write all data directly into output buffer

kristopherbullinger:
This update will restructure the SVGBackend to write all tag attribute values such as for `width`, `dx`, `viewBox` etc directly into the backend's string buffer, rather than creating temporary strings using the `format!` macro.

The existing workflow for opening a tag includes providing a tag type, a list of attrs, and whether the tag is self-closing or not. Because the attrs are provided in a list, all list items' values must have a common type of `&str`. As such, attrs which exist in the code in the form of integers or any other non-string type must first be written to a string before being included in the attrs list.

To avoid the cost of allocating many temporary strings, we instead use an `AttrWriter`. When `SVGBackend::open_tag` is called, you are given an `AttrWriter`, which includes methods for writing a key followed by a value. Typestate is used to ensure that a value is only written if there is an associated key. The `AttrWriter::write_value` method is generic over its parameter so that each type can decide how it is written to the buffer: string-like types utilize the existing html-escaping functionality, while number-like types are written directly into the buffer with their Display impl. When the user has finished writing all attrs to the tag, they will then call `AttrWriter::close` to finish a self-closing element with `/>`, or `AttrWriter::finish_without_closing` to finish the opening tag with `>` and then push the tag onto the tag_stack  so that it can be closed in the future. Thus, it preserves existing behavior.

### DIFF
--- a/plotters-svg/src/svg.rs
+++ b/plotters-svg/src/svg.rs
@@ -589,17 +589,17 @@ impl<'a> DrawingBackend for SVGBackend<'a> {
             FontTransform::Rotate90 => {
                 attrwriter
                     .write_key("transform")
-                    .write_value(("rotate(90,", x0, ',', y0, ')'));
+                    .write_value(("rotate(90, ", x0, ", ", y0, ')'));
             }
             FontTransform::Rotate180 => {
                 attrwriter
                     .write_key("transform")
-                    .write_value(("rotate(180,", x0, ',', y0, ')'));
+                    .write_value(("rotate(180, ", x0, ", ", y0, ')'));
             }
             FontTransform::Rotate270 => {
                 attrwriter
                     .write_key("transform")
-                    .write_value(("rotate(270,", x0, ',', y0, ')'));
+                    .write_value(("rotate(270, ", x0, ", ", y0, ')'));
             }
             _ => {}
         }
--- a/plotters-svg/src/svg.rs
+++ b/plotters-svg/src/svg.rs
@@ -443,7 +443,9 @@ impl<'a> DrawingBackend for SVGBackend<'a> {
             .write_value(style.stroke_width());
         attrwriter
             .write_key("points")
-            .write_value(FormatEscapedIter(path.into_iter().map(|c| (c.0, ',', c.1))));
+            .write_value(FormatEscapedIter(
+                path.into_iter().map(|c| (c.0, ',', c.1, ' ')),
+            ));
         attrwriter.close();
         Ok(())
     }
@@ -465,7 +467,9 @@ impl<'a> DrawingBackend for SVGBackend<'a> {
             .write_value(make_svg_color(style.color()));
         attrwriter
             .write_key("points")
-            .write_value(FormatEscapedIter(path.into_iter()));
+            .write_value(FormatEscapedIter(
+                path.into_iter().map(|c| (c.0, ',', c.1, ' ')),
+            ));
         attrwriter.close();
 
         Ok(())
--- a/plotters-svg/src/svg.rs
+++ b/plotters-svg/src/svg.rs
@@ -15,13 +15,9 @@ use std::io::Cursor;
 use std::io::{BufWriter, Error, Write};
 use std::path::Path;
 
-fn make_svg_color(color: BackendColor) -> String {
-    let (r, g, b) = color.rgb;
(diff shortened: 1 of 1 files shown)


plotters-svg/src/svg.rs
```diff
@@ -80,36 +77,162 @@ pub struct SVGBackend<'a> {
...
+
+impl<'a> SVGBackend<'a> {
+    fn escape_and_push(buf: &mut String, value: &str) {
+        value.chars().for_each(|c| match c {
+            '<' => buf.push_str("&lt;"),
+            '>' => buf.push_str("&gt;"),
+            '&' => buf.push_str("&amp;"),
+            '"' => buf.push_str("&quot;"),
+            '\'' => buf.push_str("&apos;"),
+            other => buf.push(other),
```
AaronErhardt, line 232: This is duplicated code, you could reuse the implementation of `FormatEscaped` for `char`

plotters-svg/src/svg.rs
```diff
@@ -222,19 +350,17 @@ impl<'a> DrawingBackend for SVGBackend<'a> {
...
+        let mut attrwriter = self.open_tag(SVGTag::Rectangle);
+        attrwriter.write_key("x").write_value(point.0);
+        attrwriter.write_key("y").write_value(point.1);
+        attrwriter.write_key("width").write_value("1");
+        attrwriter.write_key("height").write_value("1");
+        attrwriter.write_key("stroke").write_value("none");
+        attrwriter
+            .write_key("opacity")
+            .write_value(make_svg_color(color));
+        attrwriter.write_key("fill").write_value(color.alpha);
```
AaronErhardt, line 362: I think opacity and color are accidentally mixed here.

AaronErhardt (changes requested):
Code looks good overall, but there are a couple of new clippy warnings, code duplication and correctness issues. Once those are addressed, I'll do another review.

kristopherbullinger:
Good catch. Thanks. Code updated.

plotters-svg/src/svg.rs
```diff
@@ -333,21 +456,18 @@ impl<'a> DrawingBackend for SVGBackend<'a> {
...
+        let mut attrwriter = self.open_tag(SVGTag::Polygon);
+        attrwriter
+            .write_key("opacity")
+            .write_value(style.color().alpha);
+        attrwriter
+            .write_key("fill")
+            .write_value(make_svg_color(style.color()));
+        attrwriter
+            .write_key("points")
+            .write_value(FormatEscapedIter(path.into_iter()));
```
AaronErhardt, line 468: This can't be correct given that 446 does it differently. Plus in line 446 there might be a whitespace missing at the end of each point.

AaronErhardt (changes requested):
I just found one more thing.

Also, would you mind running a couple of tests to verify that the generated SVG does work as before, at least for the examples? With changes like this it's easy to overlook something and I don't want to deal with avoidable regressions.

kristopherbullinger:
You were right about the missing spaces between line points. I've just pushed a fix for that. This time I compared the files generated by the tests byte-for-byte against those generated by `7e7d2aa463a202a8ff1b340f41feb703482e753a` (from which my repo is forked) to confirm they are all equal:
```
cmp target/test/svg_master/test_draw_mesh_negative_ticks.svg target/test/svg/test_draw_mesh_negative_ticks.svg
cmp target/test/svg_master/test_draw_mesh_no_ticks.svg target/test/svg/test_draw_mesh_no_ticks.svg
cmp target/test/svg_master/test_draw_pixel_alphas.svg target/test/svg/test_draw_pixel_alphas.svg
cmp target/test/svg_master/test_series_labels.svg target/test/svg/test_series_labels.svg
cmp target/test/svg_master/test_text_alignments.svg target/test/svg/test_text_alignments.svg
cmp target/test/svg_master/test_text_clipping.svg target/test/svg/test_text_clipping.svg
cmp target/test/svg_master/test_text_draw.svg target/test/svg/test_text_draw.svg
```
returns no errors. Also, as a sanity check, I manually inspected each line side-by-side and confirmed visually that the outputs are equal. 

Thanks for your careful review. Next time I will be more thorough to ensure behavior is unchanged.

AaronErhardt (approved):
Thanks for your contribution and for checking against the master branch!

PR #638, opened 2024-09-13
Fix backend example in README

A-K-O-R-A:
In the Section "Drawing Backends" under "Concepts by example", the commented line for the SVGBackend is missing a `mut` keyword, this PR resolves this issue.


AaronErhardt (approved):
Thanks!

PR #643, opened 2024-09-26
Fix typo: cubiod -> cuboid

fotonick:
Ultra-simple search-and-replace typo fix. `cargo test` passes. No benchmark changes. No doc layout changes. Link from contributors guide to [plotters/examples/wasm-demo](https://github.com/plotters-rs/plotters/tree/master/plotters/examples/wasm-demo) is broken, but there's hardly any room for a search-and-replace to break it, so I request a waiver.


AaronErhardt:
As simple as this change might seem, I can't merge this currently because on the main branch we don't accept breaking changes. Only on `next-release-devel` we do.

fotonick:
I see. I didn't know about `next-release-dev` since it wasn't mentioned in CONTRIBUTING.md. Looking at this branch, I see that this identical patch was committed two years ago in c8364744a9d6bfc4e9c2128189a98ea1a292b494. I'm a little concerned that it hasn't made its way to `main` in that time. Either way, this PR is redundant. Closing.

Issue #644, opened 2024-09-26
[BUG] Broken link to WASM test instructions in CONTRIBUTING.md


fotonick:
**Describe the bug**
[CONTRIBUTING.md](https://github.com/plotters-rs/plotters/blob/master/CONTRIBUTING.md) contains a link to [plotters/examples/wasm-demo](https://github.com/plotters-rs/plotters/tree/master/plotters/examples/wasm-demo), which is broken (404).

**To Reproduce**
Browse to [CONTRIBUTING.md](https://github.com/plotters-rs/plotters/blob/master/CONTRIBUTING.md) and clink on the link with text "plotters/examples/wasm-demo".

**Version Information**
Accurate as of `1f174c8b8e4b9dbcb160536228510515031d4609`.

domlupo:
The source code was fixed by https://github.com/plotters-rs/plotters-wasm-demo/commit/37f929f473b3b69ebe9bd4aa88c422a475dfcba7 but it needs deployed.

PR #645, opened 2024-09-26
Fix "histogram" feature to be all_series rather than all_elements

fotonick:
I noticed that the `histogram` feature was bundled with `all_elements`, when it should be `all_series`, so I fixed it. I don't see any tests or CI configurations that would change based on this. With this patch on a local repo, I tested that one can update the histogram example in [plotters-doc-data/evcxr-jupyter-integration.ipynb](https://github.com/plotters-rs/plotters-doc-data/blob/0d6c742decbebed5b92bd85c6defa279da48cbef/evcxr-jupyter-integration.ipynb) to remove the `all_elements` feature and start working correctly.

Tests pass. Benchmarks unchanged. Doc formatting doesn't change. WASM untested due to broken link to instructions (#644), but it shouldn't change at all.


AaronErhardt:
Same here, that could potentially break the code of a user, so I'm not going to merge this into main. I hope you understand.

fotonick:
Oy, I tried the Github feature of changing the merge target, but it's all a garbled mess now. Will close now and reopen when I've properly rebased to `next-release-devel`.

PR #646, opened 2024-09-27
Fix histogram feature grouping on next-release-devel

fotonick:
Same goal as before, but since it is an API-breaking change, it is now rebased to `next-release-devel`. Note that I also had to fix a `Pie::new()` doctest that failed to compile.

### DIFF
--- a/plotters/Cargo.toml
+++ b/plotters/Cargo.toml
@@ -66,8 +66,8 @@ default = [
         "deprecated_items",  "all_series", "all_elements",
         "full_palette"
 ]
-all_series = ["area_series", "line_series", "point_series", "surface_series"]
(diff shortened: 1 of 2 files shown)


fotonick:
@AaronErhardt I fixed the CI build on this branch, in addition to my original patch. I hope it reduces chaos just a little for you.

If it's further helpful, I can commit a `Cross.toml` to aid multi-arch testing. From my Mac, I was able to `cross test --target x86_64-unknown-linux-gnu` to reproduce and debug the CI failures.

fotonick:
You're quite right that the last patch to `datetime.rs` (pulling from master) was a big step backwards. Sorry about that. This should be better. I think it leaves the codebase better than how I found it.

PR #659, opened 2024-11-01
style/color: derive serialize & deserialize

10ne1:
Using serde is useful in production, not just during development, so move it to optional dependencies instead of dev-dependencies.

This allows to derive serialize/deserialize for the color structs which are very simple using the default implementation, allowing to pass colors along via serialization channels and formats.

### DIFF
--- a/README.md
+++ b/README.md
@@ -571,6 +571,22 @@ pub fn register_font(
 ) -> Result<(), InvalidFont>
 ```
 
+- (De)serialization features
+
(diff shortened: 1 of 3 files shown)


AaronErhardt (changes requested):
I totally understand the intend of the PR and I agree that this is useful. However, some might want to use plotters without pulling in serde as dependency. Therefore, this feature should be added behind a feature flag IMO.

AaronErhardt (approved):
Thanks, looks good!

PR #663, opened 2024-12-10
Fix: `histogram` data draw

mrxiaozhuox:


mrxiaozhuox:
Sorry, I figured it out. The original code doesn't seem to have any issues.

PR #664, opened 2024-12-19
Estimate Text Layout using an OffscreenCanvas

jwagner:
I tried to use plotters in a web worker and ran into a panic in estimate_layout. The reason is fairly obvious, there is no window or document inside of a worker context. 

To get plotters to work inside of a web worker I hacked up some code to measure the text size using an [OffscreenCanvas](https://developer.mozilla.org/en-US/docs/Web/API/OffscreenCanvas).

There are some benefits and drawbacks to this. 

I think it should be more resilient than creating a span in the dom and measuring offsetWidth/Height because it is not affected by CSS.

It should be more performant than adding a dom node and measuring the offset properties which [cause layout reflows](https://gist.github.com/paulirish/5d52fb081b3570c81e3a). This is just an educated guess. Didn't benchmark it - takes this claim with a pinch of salt.

I think the main drawback is compatibility. While it is considered to be baseline now, [browser support isn't ideal](https://caniuse.com/?search=offscreencanvas).

There are several ways I see forward. This approach could be used as a fallback when window isn't available or visa versa. 

I think it's also an option to just fall back to a really crude approximation along the lines of height = size, width = size * text.len()


AaronErhardt:
Sounds reasonable to me, but I'm not proficient enough to tell which approach is preferable. In the end, both solution seem like a hack to me, and the question is which hack has the better trade-offs. I'd like to hear another opinion on this matter, but unfortunately I'm the only somewhat active maintainer currently.

jwagner:
Thanks a lot for getting back to me and keeping the project afloat @AaronErhardt. 

If browser compatibility is the main concern I could also just add a fallback to a regular (non-offscreen) canvas. That way there is (essentially) only one code path and the dom hackery can go. Canvas is supported in [practically every browser](https://caniuse.com/?search=canvas) that [supports wasm](https://caniuse.com/?search=wasm).

philocalyst:
Sure, man, I'm just helping out around the repo and know this would make it more maintainable and more likely to be considered. Do with that what you will...

▰▰▰▰▰
Miles Wirht 🙃

AaronErhardt:
> @philocalyst I'm happy to add the fallback as soon as someone with merge permissions confirms that this is what they want.

I would merge it :)

jwagner:
> > @philocalyst I'm happy to add the fallback as soon as someone with merge permissions confirms that this is what they want.
> 
> I would merge it :)

Awesome. I'm currently injured and limited in how much I can type but I will get back to this!

PR #673, opened 2025-02-12
Add ReversibleRanged support for LogCoord.

booti386:
This allows to use ChartContext::into_coord_trans() on a chart with a logarithmic axis.


booti386:
@AaronErhardt, just a single ping, so it doesn't get lost in limbo

AaronErhardt (approved):
Looks good, sorry for the late review.

booti386:
Seems like i forgot to commit the hunk at the top of the file... Should be ok now.

AaronErhardt (approved):
Thanks!

PR #694, opened 2025-05-06
Exposing LogRangeExt

maxime-bruno:
LogRangeExt was not exposed and as such, users could not define their custom log range


philocalyst:
Bro is right @AaronErhardt

PR #704, opened 2025-10-24
For the DualCoordChartContext add borrow_secondary_mut function

bennysj:
The function borrow_secondary function does not return a mutable reference. It would be useful if it actually would be possible to get a &mut reference to the secondary context.
Implement a new function that actually return a &mut reference to the secondary context.

For example to write a generic draw function:
```rust 
    fn draw_chart_series<DB>(
        chart: &mut ChartContext<'_, DB, Cartesian2d<RangedCoordf32, RangedCoordf32>>,
        data_series: &ModelRc<XYData>,
    ) where
        DB: DrawingBackend,
    {
        // draw the chart series for both primary and secondary
   }
```


nielsle:
This looks useful.

PR #710, opened 2026-01-02
correct links to backends in main module description

jonaspleyer:


philocalyst:
I would move for this to be closed @AaronErhardt 

Inactivity, and small to begin with.

PR #711, opened 2026-01-04
add support for disabling anti-aliasing in BitmapBackend

jonaspleyer:
## Description
This PR arose from a personal need where I wanted to quickly draw some geometries but they had to be solid of identical color. In this scenario, anti-aliasing is actually undesirable. Since I saw no functionality to disable this, I added it.

The PR does not yet contain any tests. I would love to hear some feedback if this approach is reasonable.
The introduction of a new defaulted method `fn use_anti_aliasing(&self) -> bool` to the `DrawingBackend` trait would probably be classified as a minor change by Rust's semver-specs https://doc.rust-lang.org/cargo/reference/semver.html#trait-new-parameter-default thus requiring a minor version bump. However, I am happy to hear any feedback on this.

## Example Usage
```rust
let (x, y) = ..;
let mut buffer = vec![0u8; x*y*3];
BitMapBackend::with_buffer(&mut buffer, (x, y))
    .anti_aliasing(false)
    .into_drawing_area();
```

### DIFF
--- a/plotters-backend/src/lib.rs
+++ b/plotters-backend/src/lib.rs
@@ -326,4 +326,10 @@ pub trait DrawingBackend: Sized {
 
         Ok(())
     }
+
+    /// Utilize anti-aliasing if supported by the chosen backend.
+    /// By default this option is true.
+    fn use_anti_aliasing(&self) -> bool {
+        true
+    }
 }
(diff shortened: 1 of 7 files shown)


philocalyst:
@jonaspleyer The approach looks good! The project is in a maintenance state right now, so tests would be appreciated to lower a contribution to the sum burden!

(not a maintainer, just aware)

philocalyst:
Following up @jonaspleyer

jonaspleyer:
I will check if I can find the time to write some tests. However I am currently very busy and so it might take a while.

PR #717, opened 2026-01-23
Fix: negative tick_mark_size now only affects tick direction, not labels

### LINKED ISSUE #716: [BUG] Negative tick_mark_size moves labels into plot area instead of only changing tick direction
**Describe the bug**

When using `set_tick_mark_size` with negative values, the intention is to make only the tick marks point inward (into the plot area), which is a common style in scientific papers. However, three problems occur:

1. **Labels move with tick marks**: The numeric tick labels and axis descriptions also move into the plot area, instead of staying in their original positions outside the plot.

2. **Axis description overlaps with numeric labels**: Because both move inward, the y-axis description ("y (nondimensional)") overlaps with the numeric tick labels.

3. **White background hides grid lines**: The moved labels have a white background that covers the grid lines behind them.

**Expected behavior**:
- Only the tick marks should point inward
- Numeric labels and axis descriptions should remain outside the plot area (their normal positions)
- Similar to matplotlib's `ax.tick_params(direction='in')`

**Actual behavior**:
- Tick marks, numeric labels, and axis descriptions all move into the plot area
- Labels overlap each other
- Grid lines are hidden by white label backgrounds

Image

**To Reproduce**

```rust
use plotters::prelude::*;

    let root = BitMapBackend::new("inward_ticks.png", (800, 600)).into_drawing_area();
    root.fill(&WHITE)?;

    let mut chart = ChartBuilder::on(&root)
        .margin(10)
        .x_label_area_size(40)
        .y_label_area_size(60)
        .build_cartesian_2d(-4.0f32..4.0f32, -4.0f32..4.0f32)?;

    chart
        .configure_mesh()
        .x_desc("x (nondimensional)")
        .y_desc("y (nondimensional)")
        .set_tick_mark_size(LabelAreaPosition::Left, -5)
        .set_tick_mark_size(LabelAreaPosition::Bottom, -5)
        .draw()?;

    root.present()?;
    Ok(())
}
```

**Version Information**

- plotters: 0.3.7 (from crates.io)
- rustc: 1.87.0 (17067e9ac 2025-05-09)
- OS: Ubuntu 24.04.3 LTS

chantakan:
## Summary
This PR fixes the behavior of negative `tick_mark_size` values. Previously, using a negative value caused labels and axis descriptions to move inside the plot area, overlapping with the chart content. 

## Expected Behavior (after this fix)
When using negative `tick_mark_size`, only the tick marks should point inward (into the plot area), while labels remain outside. This matches the expected behavior similar to matplotlib's `ax.tick_params(direction='in')`.

## Previous Behavior (bug)
- Labels moved into the plot area
- Axis descriptions overlapped with numeric labels
- White label backgrounds hid grid lines

## Changes
- Labels are now always positioned outside the plot area regardless of tick direction
- Axis lines are always drawn on the outer edge
- Inward tick marks are drawn on the plot area using `strip_coord_spec()`
- Outward tick marks continue to work as before

## Testing
- All 81 unit tests pass
- All 49 doc tests pass
- Visually verified with test examples

Fixes #716

### DIFF
--- a/plotters/src/chart/context/cartesian2d/draw_impl.rs
+++ b/plotters/src/chart/context/cartesian2d/draw_impl.rs
@@ -153,7 +153,8 @@ impl<'a, DB: DrawingBackend, X: Ranged, Y: Ranged> ChartContext<'a, DB, Cartesia
 
         /* Draw the axis and get the axis range so that we can do further label
          * and tick mark drawing */
-        let axis_range = self.draw_axis(area, axis_style, orientation, tick_size < 0)?;
+        // Always draw axis on the outer edge (pass false for inward_labels)
+        let axis_range = self.draw_axis(area, axis_style, orientation, false)?;
 
         /* To make the right label area looks nice, it's a little bit tricky, since for a that is
          * very long, we actually prefer left alignment instead of right alignment.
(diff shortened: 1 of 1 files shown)


philocalyst:
We need your workflow approval @AaronErhardt but I ran this on my machine and everything seems to be swell! Code is also good !

PR #721, opened 2026-02-11
feat: add tooltip support

### LINKED ISSUE #13: [feature request] Interactivity
Unless I missed it, there is no way to create interactive plots for wasm or native.

80avin:
fixes https://github.com/plotters-rs/plotters/issues/13
It works, but I don't like the presence of
* DrawingBackend::with_tooltips function
   tooltip should be enabled per chart, not backend.
* `draw_series_with_tooltips` api.
  Ideally, it should work with draw_series API and tooltip should be a configuration parameter.

Challenge:
* The library gives user the control over when to draw. Because of this, we don't have enough information in draw_series to emit the tooltips code.

Alternative API:
* `chart.tooltip_mode(TooltipMode)` shall enable tooltips for current chart.
   mode can be "Auto", "Always", "Disable". If Always and the backend doesn't support, there should be error.
* On every draw ( and .label, etc) call, keep on building internal state.
   Also, SVG backend shall add classes wherever required to uniquely identify elements, if tooltip mode is enabled.
* On present call, add javascript to enable tooltips.

Question to the maintainers:
* Does the API feel satisfactory, or do you have better suggestions?

image

AI use:
The initial approach, including the API was designed after manual experiments. AI is used to scale the approach into a wider & cleaner solution.

### DIFF
--- a/plotters-backend/src/lib.rs
+++ b/plotters-backend/src/lib.rs
@@ -76,6 +76,92 @@ use text_anchor::{HPos, VPos};
 /// which defines the top-left point as (0, 0).
 pub type BackendCoord = (i32, i32);
 
+/// Describes how pixel positions along an axis map to logical valeus.
+///
+/// Backends can use this to compute tooltip labels for arbitrary cursor positions (continuous) or
+/// to snap to the nearest labeled tick (discrete).
+#[derive(Clone, Debug, PartialEq)]
+pub enum Interpolation {
+    /// A linear mapping from a pixel range to a floating-point value range.
+    Continuous {
+        /// The pixel range that this axis occupies on the backend.
+        backend_range: (i32, i32),
+        /// The logical value range corresponding to `backend_range`.
+        value_range: (f64, f64),
+        /// Human-readable units string (may be empty).
+        units: String,
+    },
+    /// A set of individually labeled positions (tick marks).
+    Discrete {
+        /// Each tuple is (pixel_position, label_text).
+        points: Vec<(i32, String)>,
+    },
+}
+
+/// Semantic context passed to [`DrawingBackend::begin_context`] so that backends know *what* is
+/// being drawn and can attach metadata such as tooltip information, accessibility labels, or
+/// interactive behaviour.
+///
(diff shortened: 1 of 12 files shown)


philocalyst:
Still working on this? @80avin

80avin:
Waiting for opinions on the architecture.
The PR works for most cases already.

philocalyst:
If you're using AI for the PR, you need to signal it somewhere, and ideally link to the thread.

Also, please use the [SVG](https://lib.rs/crates/svg) crate rather than brittle raw strings.

Just starting advice. 

▰▰▰▰▰
Miles Wirth 🙃

80avin:
Thanks Miles.
I added the note about AI usage in the PR description.

Also, phrased a clearer question to maintainers in the PR description. Completing the PR is not an issue, but I am myself unconvinced by the current API. I went with this API design while trying to keep the breaking changes to a minimum; otherwise, the backend system would have needed major breaking changes

philocalyst:
Breaking changes like what?

(Kind of a psuedo-maintainer)

▰▰▰▰▰
Miles Wirth 🙃

80avin:
I've tried to explain that in the PR description. If something is difficult to understand, I can take specific questions.

In abstract terms, breaking changes like
* Changing what data is shared with backends.
   Currently, backends only get low level information of what figure to draw and where. But, for a backend to draw tooltips or other decorations, it needs to have some semantic information about the chart.
   The `begin_context`/`end_context` that I added work but are after-thoughts, designed to add tooltip support without changing much in existing design.
* Changing the user facing API.
  As I mentioned in the PR description, the backend.with_tooltip is not ideal. Ideally, we would want to enable/disable tooltip per chart ( or element of chart ) than entire backend.

Let me know your thoughts on these, or if I'm thinking too much. If the current design looks good, I can finish the PR and move out of draft.

philocalyst:
I think that a Context object, where start and end are fields, and further information could be added to would be the best step forward. It should have a default implementation, and notably not optional (Particular backends should make it explicit when they're skipping out on the object).

▰▰▰▰▰
Miles Wirth 🙃

PR #735, opened 2026-04-25
Remove `font-kit` and `pathfinder` as dependencies

pbdeuchler:
The servo organization no longer maintains `font-kit` or `pathfinder` and has only been doing ad hoc PR fixes for failing builds. However as Rust continues to change things in the `simd` std lib crate `pathfinder` continues to break builds and it has been communicated that we should not rely on releases happening in the future. 

This PR aims to completely remove `font-kit` and `pathfinder` dependencies in favor of the maintained `fontique` and `swash` crates. It uses `fontique` to discover system fonts and build an internal cache of available fonts, and `swash` for rendering. 

### Tradeoffs

We do lose subpixel AA rendering with this PR, however the introduction of `swash` will allow us to improve typographic quality with a future PR (which I already have in progress) that removes `ttf-parser` in favor of a 100% `swash` rendering backend, which would also mean improved support for ligatures and non-Latin fonts. `swash` also does slightly more allocations and is slightly less performant than `pathfinder`, but the perf difference here is extremely small and I doubt anyone will notice in real world usage.

### Breaking Changes

I changed the error type in ttf.rs since the underlying types no longer apply, this could break some public API consumers

This PR will also up the MSRV to 1.88 and introduces `memmap2` as a transitive dependency of `fontique`.

This is my first foray into the plotters internals, and my first time messing around with fonts/rasterization... I utilized my good friend Claude to guide me but let me know if there's anything that's missing or needs fixing here.

### DIFF
--- a/.github/workflows/plotters-backend.yml
+++ b/.github/workflows/plotters-backend.yml
@@ -12,6 +12,9 @@ jobs:
     - uses: actions/checkout@v4
       with:
           submodules: recursive
+    - name: Install fontconfig
+      if: runner.os == 'Linux'
+      run: sudo apt-get update && sudo apt-get install -y fontconfig
     - uses: actions-rs/toolchain@v1
       with:
           toolchain: stable
--- a/.github/workflows/plotters-backend.yml
+++ b/.github/workflows/plotters-backend.yml
@@ -14,7 +14,7 @@ jobs:
           submodules: recursive
     - name: Install fontconfig
       if: runner.os == 'Linux'
-      run: sudo apt-get update && sudo apt-get install -y fontconfig
+      run: sudo apt-get update && sudo apt-get install -y libfontconfig-dev
     - uses: actions-rs/toolchain@v1
       with:
           toolchain: stable
--- a/.github/workflows/plotters-bitmap.yml
+++ b/.github/workflows/plotters-bitmap.yml
@@ -12,6 +12,9 @@ jobs:
     - uses: actions/checkout@v4
       with:
           submodules: recursive
+    - name: Install fontconfig
+      if: runner.os == 'Linux'
+      run: sudo apt-get update && sudo apt-get install -y fontconfig
     - uses: actions-rs/toolchain@v1
       with:
           toolchain: stable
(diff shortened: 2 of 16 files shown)


pbdeuchler:
`fontique` requires `libfontconfig-dev` to be installed on build, which is a bit unfortunate but I've added that to the github workflows

pbdeuchler:
@philocalyst @AaronErhardt  let me know your thoughts on this, and if this is something you all are interested in. Would love to get plotters building on nightly without worry again. Thank you!

philocalyst:
Thanks so much for tagging me! Still not a maintainer so I don't see as much of these as I wish to --  when it comes to Rust and the font stack there's a lot of movement still going on. I believe for rendering the main swash maintainer has actually been hired by the google fonts team -- so I think we should be using harfrust instead.

pbdeuchler:
man the rust ecosystem drives me absolutely nuts sometimes... that sounds fine to me, i'll take a look at that and get back to you

philocalyst:
Thanks so much! I'll do another review when that's swapped, sorry for the duplicate work :(

Tests are definitely appreciated, good work there from my skim.

pbdeuchler:
No worries. The more I look into this I wonder if there's a better approach. How do you feel about me refactoring things so that font lookup happens at the `draw()` call, and I can make the choice to not use system fonts a programmatic option. Then we can just create one font rendering path and remove the `ab_glyph` dependency/feature. 

I'm pretty sure I can make this a backwards compatible change as well so we can just deprecate the existing feature/register function without removing it.

philocalyst:
That sounds perfect! Removing ab_glyph especially, it's relatively unmaintained. 

▰▰▰▰▰
Miles Wirth 🙃

nicoburns:
> Thanks so much for tagging me! Still not a maintainer so I don't see as much of these as I wish to -- when it comes to Rust and the font stack there's a lot of movement still going on. I believe for rendering the main swash maintainer has actually been hired by the google fonts team -- so I think we should be using harfrust instead.

Swash is still the recommended solution for rasterizing glyphs atm.

[Glifo](https://github.com/linebender/vello/tree/main/glifo) is being developed as a replacement based on the "fontations" stack from Google Fonts, but it isn't ready yet. HarfRust does shaping, not rendering.

PR #736, opened 2026-04-29
Refactor font rendering pipeline, remove `font-kit` and `pathfinder`

pbdeuchler:
Alternative PR to #735 

## Font rendering pipeline changes

Per discussion [here](https://github.com/plotters-rs/plotters/pull/735#issuecomment-4340664968), with the removal of `font-kit` and `pathfinder` we are now able to unify the default font rendering pipeline into a single code path. Users can now determine whether or not system fonts are used programmatically as part of the API instead of using build flags. The biggest change outside of the API surface is that font resolution and rasterization is deferred until the chart's `draw` call actually runs.

The new API creates a `FontContext` struct that is created and scoped per `DrawingArea`. This `FontContext` struct gates what fonts are available for `draw` calls. Users may manipulate the `FontContext` by calling the new `DrawingArea::with_fonts` method. This call differs only slightly from the existing `register_font` call in that it takes multiple fonts and relaxes the `static` requirement on the font bytes by allowing `Arc<[u8]>`. This obviously creates some allocations but I'm not sure that's a concern here. Happy to change that back or into something different if we'd prefer a different performance profile.

## Deprecation of `ab_glyph` feature/functionality

With this change we can deprecate the `ab_glyph` feature and `register_font` functionality. The only thing the `ab_glyph` feature enables now is:
1. The `register_font` symbol and its global registry exist.
2. The default `FontContext` is built with `enable_system: false`.

I've done my best to integrate the existing behavior into the new pipeline however due to the way font caching works now `register_font` is much more stateful than before. There is a potential footgun when using both `register_font` and `with_fonts` where `register_font` can override the global state and unintentionally change a specific `DrawingArea`'s `FontContext`, however this behavior works in an obvious imperative way so I'm hoping it's not a big issue.

In order to make `register_font` work nicely with the new caching there are some performance changes where on every text render call we now:

- Walk explicit Vec (O(N) where N is small, typically ≤ 5)
- Maybe clone the registry Vec and walk it (using `ab_glyph`)
- Maybe one fontique system query (cached internally by fontique's Collection + SourceCache)
- Hash the bytes and look up GLOBAL_PARSED (typical case: hit, returns cached Arc)

Parsing is still amortized and even for programs where users are using hundreds of fonts with hundreds of charts there should be a negligible difference, however anything rendering dynamically changing charts at scale will definitely notice something.

## Removal of `font-kit` and `pathfinder`

This PR also removes `font-kit` and `pathfinder` in favor of `harfrust`, `skrifa`, and `zeno`. It also creates a `FontEngine` abstraction so this trio of dependencies can easily be swapped out in the future as things change (or to `swash` if we want). This is probably the most straightforward change. 

There is a potential regression since `register_font` takes byte blobs where existing code that is compatible with `ab_glyph` fails with `harfrust`. However given that `harfrust` is a Google sanctioned port of `harfbuzz`, the industry gold standard of font parsing, if there are regressions they are probably edge cases that shouldn't exist in the first place.

We also bump MSRV to 1.88.0.


pbdeuchler:
@philocalyst let me know your thoughts on this one... completely open to feedback, i'm not super attached to how I implemented things here but I think it's the least worst option. I might still refine this a bit and have some more manual testing to do but let me know your initial thoughts.

pbdeuchler:
Hmmmm, I'll take a look at those CI failures in a bit. I think I know what's happening.

doc-template/readme.template.md
```diff
@@ -270,25 +270,30 @@ The following list is a complete list of features that can be opted in or out.
...
-This function only exists when the `ab_glyph` feature is enabled.
+Native text rendering is always available. Plotters resolves system fonts
+through `fontique`, shapes text with `harfrust`, reads outlines with `skrifa`,
+and rasterizes glyphs with `zeno`. Use `DrawingArea::with_fonts` when a chart
+should use in-memory fonts instead of, or in addition to, system fonts. See
+[`plotters/examples/dynamic_font.rs`](plotters/examples/dynamic_font.rs) for an
+end-to-end example that downloads Roboto from Google Fonts at runtime and
+attaches it to a chart through the new pipeline.
+
+| Name     | Description                                                                       | Additional Dependency | Default? |
```
philocalyst, line 287: ab_glyph doesn't add the ab_glyph dep?

plotters/examples/dynamic_font.rs
```diff
@@ -0,0 +1,145 @@
...
+use plotters::prelude::*;
+use std::error::Error;
+use std::io::Read;
+use std::sync::Arc;
+
+const FONTS_CSS_URL: &str = "https://fonts.googleapis.com/css2\
+    ?family=Kablammo\
+    &family=Roboto:ital,wght@0,100..900;1,100..900\
+    &display=swap";
+
```
philocalyst: Would rather just host the fonts in-tree, to reduce this logic and avoid the issues when polling against an external service.

plotters/src/style/font/engine.rs
```diff
@@ -0,0 +1,102 @@
...
+    /// Rasterize a single glyph at the requested pixel size.
+    ///
+    /// `subpixel` carries the fractional `(x, y)` offset of the glyph origin
+    /// within its target pixel cell, in `[0, 1)`. Implementations should fold
+    /// it into the rasterization so strokes that fall between integer pixel
+    /// columns are anti-aliased correctly instead of being rounded away.
+    fn rasterize(
+        &self,
+        glyph_id: u32,
+        size_px: f32,
```
philocalyst: What is size_px for? Feel like the name is washy

plotters/src/style/font/harfrust_engine.rs
```diff
@@ -0,0 +1,292 @@
...
+    glyph
+        .draw(
+            DrawSettings::unhinted(Size::new(size_px), LocationRef::default()),
+            &mut ZenoPen {
+                path,
+                sx: subpixel.0,
+                sy: subpixel.1,
+            },
+        )
+        .map(|_| ())
```
philocalyst, line 217: Can remove the map (_) call and use `?` at the end of the map_err

plotters/src/style/font/harfrust_engine.rs
```diff
@@ -0,0 +1,292 @@
...
+        let font = self.skrifa_font()?;
+        let metrics = font.metrics(Size::new(size_px), LocationRef::default());
+        let min_y = (-metrics.ascent).floor() as i32;
+        let max_y = (-metrics.descent).ceil() as i32;
+        let max_y = if max_y > min_y {
+            max_y
+        } else {
+            size_px.ceil() as i32
+        };
+        let width = (cursor_x * scale).ceil().max(0.0) as i32;
```
philocalyst, line 143: Can be simplified for readability
```suggestion
let font = self.skrifa_font()?;
let metrics = font.metrics(Size::new(size_px), LocationRef::default());
let min_y = (-metrics.ascent).floor() as i32;
let descent_y = (-metrics.descent).ceil() as i32;
let max_y = if descent_y > min_y { descent_y } else { size_px.ceil() as i32 };
let width = (cursor_x * scale).ceil().max(0.0) as i32;
```

plotters/src/style/font/harfrust_engine.rs
```diff
@@ -0,0 +1,292 @@
...
+        let key = size_px.to_bits();
+        if let Some(h) = self
+            .hinters
+            .lock()
+            .map_err(|_| FontError::LockError)?
+            .get(&key)
+            .cloned()
+        {
+            return Ok(h);
+        }
```
philocalyst: Another readability thing
```suggestion
let hinter = self.hinters.lock()
    .map_err(|_| FontError::LockError)?
    .get(&key)
    .cloned();

if let Some(h) = hinter {
    return Ok(h);
}
```

plotters/src/style/font/harfrust_engine.rs
```diff
@@ -0,0 +1,292 @@
...
+        HarfrustFontRef::from_index(self.data.as_ref(), self.index)
+            .map_err(|_| FontError::InvalidFontIndex(self.index))
+    }
+
+    fn skrifa_font(&self) -> Result<SkrifaFontRef<'_>, FontError> {
+        SkrifaFontRef::from_index(self.data.as_ref(), self.index)
+            .map_err(|_| FontError::InvalidFontIndex(self.index))
+    }
+
+    fn hinter_for(&self, size_px: f32) -> Result<Option<Arc<HintingInstance>>, FontError> {
```
philocalyst: In general find the result<option<>> pattern needless -- anyway we could compress to just a Result<>? Would appreciate a comment inline if you do feel it is needed.

philocalyst (changes requested):

philocalyst:
Do you plan to fix the remaining CI errors? The tests and error variants feel solid to me, just some readability and semantic issues core to the logic that I would prefer to be untangled. 

I'd also like to mark somewhere that when [Glifio](https://github.com/linebender/vello/tree/main/glifo) stabilizes, to switch off of `zero` for _it_. 

That should solidify this stack for a while yet.

philocalyst:
Not a blocker for a merge, but if you'd like bonus points, pioneering screenshot integration testing here would be a great way to see how (if at all) this changes text outputs, and ensure future changes don't ruin this.

pbdeuchler:
@philocalyst I appreciate the review! working on fixing those CI issues and will address these comments as well

pbdeuchler:
I do want to say that right now i'm not super happy with the kerning and rendering on this branch. Here's some screenshots for comparison, let me know your thoughts:

## `master`

master - title

master branch date

## `swash` branch

swash branch - title

swash branch date

## `skifra`/`zeno` branch (this PR)

skifra:zeno title

skifra:zeno date

pbdeuchler:
## `master`

sample

sample

## `skifra`/`zeno` branch (This PR)

relative_size


sample

philocalyst:
Yeah that's a little annoying -- that's something handled by Harfrust as well :(

doc-template/readme.template.md
pbdeuchler: Not anymore, it all goes through the same pipeline. `ab_glyph` just turns off system font discovery now.

plotters/src/style/font/engine.rs
pbdeuchler: Pixel size for the font, for calculating scale and passing to the harfrust shaper

plotters/src/style/font/harfrust_engine.rs
pbdeuchler: Compiler complains about this, can't infer the types

plotters/src/style/font/harfrust_engine.rs
pbdeuchler: I flattened this to an `Option<>` instead since we want to throw away the error anyway and draw an unhinted glyph if this fails. Happy to make it a Result with an error log or something if we want though

pbdeuchler:
> I'd also like to mark somewhere that when [Glifio](https://github.com/linebender/vello/tree/main/glifo) stabilizes, to switch off of `zero` for _it_.

https://github.com/plotters-rs/plotters/pull/736/changes#diff-f071a3c388257ffa578325cc55317265fa82ec7198a05e2167f9a570520df378R10

@philocalyst I believe I addressed everything, ready for re-review

plotters/src/style/font/engine.rs
philocalyst: Non-blocking but I was hinting that it could be renamed for clarity -- font_size_px maybe?

plotters/src/style/font/migration.rs
```diff
@@ -0,0 +1,56 @@
...
+) -> Result<(), InvalidFont> {
+    let data = Arc::<[u8]>::from(bytes);
+    HarfrustEngine
+        .parse(data.clone(), 0)
+        .map_err(|_| InvalidFont { _priv: () })?;
+
+    REGISTERED_FONTS
+        .lock()
+        .map_err(|_| InvalidFont { _priv: () })?
+        .push(registered_font(name, style, data));
```
philocalyst: We're dropping the original errors here which feels unwise for something that interacts with foreign data, would prefer to be as helpful as possible (when) bad things happen.

plotters/src/style/font/system.rs
```diff
@@ -0,0 +1,99 @@
...
+                query.set_families([QueryFamily::Generic(GenericFamily::Monospace)])
+            }
+            FontFamily::Name(name) => query.set_families([QueryFamily::Named(name)]),
+        }
+        query.set_attributes(attributes(style));
+        if with_fallback {
+            // Latin script covers the ASCII/Latin-1 ranges that chart labels
+            // are overwhelmingly drawn from; fontique iterates `families`
+            // first and only consults the fallback list when nothing in
+            // `families` matched.
```
philocalyst:
```suggestion
```

Comment isn't helpful

plotters/src/style/font/harfrust_engine.rs
```diff
@@ -0,0 +1,297 @@
...
+}
+
+struct ZenoPen<'a> {
+    path: &'a mut Vec<Command>,
+    // Subpixel offset folded into the path coordinates so the rendered mask
+    // already accounts for the glyph's fractional pixel position. Without
+    // this, the caller has to round to integer pixel positions and the
+    // sub-pixel kerning information from harfrust is lost.
+    sx: f32,
+    sy: f32,
```
philocalyst:
```suggestion
    subpixel_offset: Vector2F,
```

plotters/src/style/font/harfrust_engine.rs
```diff
@@ -0,0 +1,297 @@
...
+            width: placement.width,
+            height: placement.height,
+            data,
+        })
+    }
+}
+
+fn draw_unhinted_glyph(
+    glyph: &OutlineGlyph<'_>,
+    size_px: f32,
```
philocalyst: Same comment as prior

plotters/src/style/font/harfrust_engine.rs
```diff
@@ -0,0 +1,297 @@
...
+    subpixel: (f32, f32),
```
philocalyst: Would like to mirror the x/y logic from below

plotters/src/style/font/engine.rs
```diff
@@ -0,0 +1,102 @@
...
+    FontUnavailable {
+        /// The requested family name.
+        family: String,
+        /// The requested style name.
+        style: String,
+    },
+    /// A glyph outline could not be converted into a coverage mask.
+    RasterizeError(String),
+    /// Internal font state could not be locked.
+    LockError,
```
philocalyst: Great error cases here :)

plotters/src/style/font/context.rs
```diff
@@ -0,0 +1,509 @@
...
+            if matches!(style, FontStyle::Normal) {
+                None
+            } else {
+                fonts.iter().rev().find(|font| {
+                    font.family == family.as_str()
+                        && font.style.as_str() == FontStyle::Normal.as_str()
+                })
+            }
+        })
+}
```
philocalyst, line 432:
```suggestion
fn find_registered_font<'a>(
    fonts: &'a [RegisteredFont],
    family: FontFamily<'_>,
    style: FontStyle,
) -> Option<&'a RegisteredFont> {
    let family_str = family.as_str();
    let style_str = style.as_str();

    let mut fallback = None;

    for font in fonts.iter().rev() {
        if font.family != family_str {
            continue;
        }
        if font.style.as_str() == style_str {
            return Some(font);
        }
        if fallback.is_none() && !matches!(style, FontStyle::Normal) && font.style.as_str() == FontStyle::Normal.as_str() {
            fallback = Some(font);
        }
    }

    fallback
}
```

Nice simplification

plotters/src/style/font/context.rs
```diff
@@ -0,0 +1,509 @@
...
+
+    pub(crate) fn draw<E, DrawFunc: FnMut(i32, i32, f32) -> Result<(), E>>(
+        &self,
+        family: FontFamily<'_>,
+        style: FontStyle,
+        size: f64,
+        text: &str,
+        (base_x, base_y): (i32, i32),
+        mut draw: DrawFunc,
+    ) -> FontResult<Result<(), E>> {
```
philocalyst: Could we not simplify this to just FontResult or Result? This is another anti-pattern imo.

plotters/src/style/font/harfrust_engine.rs
```diff
@@ -0,0 +1,297 @@
...
+            glyphs,
+            bounds: ((0, min_y), (width, max_y)),
+        })
+    }
+
+    fn rasterize(
+        &self,
+        glyph_id: u32,
+        size_px: f32,
+        subpixel: (f32, f32),
```
philocalyst: Same Vecto2f here

philocalyst (changes requested):

philocalyst:
A lot cleaner this time around, feeling good about this PR! 

Have no complaints after these (small) issues are handled. You over-comment in some places but it's not a point to press. Just wish that some of that explaining was moved to the complex or external stuff (Like the Zeno impls with the heavily abbreviated params). 

Testing looks green and good. Does make me think we need some kind of reporting, the lack of a warn in some of the fallbacks could lead to some annoying debugging for users.

Would love keep seeing contributions from you :)

plotters/src/style/font/engine.rs
pbdeuchler: lol once I started enumerating all of them I figured I might as well be exhaustive once I got 80% of the way there

plotters/src/style/font/context.rs
pbdeuchler: wups sorry I think this is LLM slop that I didn't catch

pbdeuchler:
Ha, you're not the first person to say that about my comments. TBH a big reason the more complex stuff isn't commented is because I still don't have a great grasp on what a lot of that is doing, it was mostly trial and error/copy pasting/Claude that got the rasterization and calculation bits working. Slowly figuring it out though.

I found another edge case with series labels last night, so I created a complex example chart to help test things and fixed the bug in the least hacky way I could. I'm still not happy with the kerning and rasterization, but I figure once we can get this merged I can start to experiment with other feature gated engines that might emphasize font quality.

I'm a bit busy today but will try and address your comments soon. Happy to help keep improving plotters, I use it extensively so there's a couple other things i'd like to fix if I can find the time.

pbdeuchler:
Ready for review @philocalyst 

modern

New example

philocalyst:
Looks wonderful!

philocalyst (approved):

philocalyst:
@AaronErhardt Any thoughts? Still don't have the perms for this, but looks swell :)

Hope there's no hard feelings about the msrv...

AaronErhardt:
> @AaronErhardt Any thoughts? Still don't have the perms for this, but looks swell :)

The problem is, that this is a breaking change and currently the master branch is more like a "stable" branch. I know, not very intuitive and I should change that. 

But even so, the problem is, that I neither have the time to dive into the code nor do I have the expertise in this area not do I use plotters at the moment. If there would be one additional independent review, I'd be confident in merging this, but then we still kinda need a readmap for a breaking release.

philocalyst:
@AaronErhardt Are there any other people like me you know of who are actively contributing but are without perms? Was thinking about forming some kind of chat to make life easier for you :)

AaronErhardt:
> @AaronErhardt Are there any other people like me you know of who are actively contributing but are without perms? Was thinking about forming some kind of chat to make life easier for you :)

There are but many don't stay long enough to take over maintenance. I don't want to make random contributors maintainers immediately and even those who stuck around for a while have left.

But you can try opening a chat. I'm actually looking forward to find capable people who are interested in this project because I just do the bare minimum to keep it somewhat alive.

pbdeuchler:
@AaronErhardt @philocalyst zulip/discord seem to be the popular picks for rust projects, but you need people to be active and engaging in discussion for them to be useful

Might be worth adding a banner to the top of the readme to announce we're looking for maintainers

philocalyst:
I made this discord!!!!! https://discord.gg/wH6whNtmfz

Happy to transfer ownership to you @AaronErhardt :) I will try to hunt down other people in PRs but would be nice if you linked it somewhere.

StefanSteiner:
Heads-up that `ttf-parser` was just flagged unmaintained ([RUSTSEC-2026-0192](https://rustsec.org/advisories/RUSTSEC-2026-0192), 2026-06-28). It's informational rather than a real vuln (the author has stepped away and there's no patched version), but it now trips `cargo deny` and `cargo audit` for everyone downstream of plotters.

The thing that makes it relevant here: the advisory covers every version of `ttf-parser`, so the version bumps in #662 and friends can't clear it. Right now both font paths end at it. The `ttf` default goes through `font-kit`, and the `ab_glyph` feature goes `ab_glyph` -> `owned_ttf_parser` -> `ttf-parser ^0.25.1`.

Moving the system path to `harfrust`/`skrifa`/`zeno` in this PR is exactly the right direction and clears it there. The one wrinkle is that `register_font` still rasterizes through `ab_glyph`, so `ttf-parser` would stick around in the tree as long as that path exists, and the advisory wouldn't fully clear until `ab_glyph` is retired too (or its rasterization rerouted through the new `FontEngine`). Is dropping the `ab_glyph`/`ttf-parser` path in scope for this PR, or is that a follow-up?

For context, we're a downstream crate ([tableau/hyper-api-rust](https://github.com/tableau/hyper-api-rust)) using plotters just for server-side PNG/SVG chart rendering. We've waived the advisory for now and are tracking this issue for the real fix. Thanks for taking this on, happy to test against the branch.

pbdeuchler:
@StefanSteiner this PR should actually completely remove `ttf-parser` from all crates in this project (correct me if I missed something), the `ab-glyph` feature is now solely for backwards compat... all font rasterization functionality now shares the same code path regardless of how the font was sourced, and there are new APIs for bypassing system font discovery and providing your own font data.

If you're willing to test the branch that would be great, I've been using this branch in production myself for almost two months now and haven't hit any major bugs. Happy to fix any bugs or refine things.

